PR #10, opened 2021-10-20
Remove temporary unit test JSON

streeve:
@jychoi-hpc pointed out that multiple processes writing to the `tmp.json` file was likely causing the semi-regular MPI failures 

This adds overloads for `run_training` and `run_predict` that directly pass the config dictionary (modified during the test instead)

Also cleans up the interface for `run_predict`: config should be a required input and `chosen_model` was unused

### DIFF
--- a/hydragnn/run_prediction.py
+++ b/hydragnn/run_prediction.py
@@ -19,7 +19,7 @@ from hydragnn.models.create import create
 from hydragnn.train.train_validate_test import test
 
 
-def run_prediction(config_file: str = "./examples/configuration.json"):
+def run_prediction(config_file: str):
 
     config = {}
     with open(config_file, "r") as f:
--- a/hydragnn/run_prediction.py
+++ b/hydragnn/run_prediction.py
@@ -19,13 +19,16 @@ from hydragnn.models.create import create
 from hydragnn.train.train_validate_test import test
 
 
-def run_prediction(config_file: str = None, chosen_model: torch.nn.Module = None):
+def run_prediction(config_file: str = "./examples/configuration.json"):
 
-    if config_file is None:
-        raise RuntimeError("No configure file provided")
+    config = {}
+    with open(config_file, "r") as f:
+        config = json.load(f)
 
-    if chosen_model is None:
-        raise RuntimeError("No model type provided")
+    run_prediction(config)
+
+
+def run_prediction(config: {}):
 
     try:
         os.environ["SERIALIZED_DATA_PATH"]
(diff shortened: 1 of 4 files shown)


pzhanggit (approved):

allaffa (approved):

streeve:
Not working yet. Need to fix the multiple overloads

hydragnn/run_prediction.py
```diff
@@ -19,13 +20,23 @@
 from hydragnn.train.train_validate_test import test
 
 
-def run_prediction(config_file: str = None, chosen_model: torch.nn.Module = None):
+@singledispatch
```
allaffa, line 23: Why do we need this?

hydragnn/run_prediction.py
```diff
@@ -19,13 +20,23 @@
...
+def run_prediction(config):
+    raise TypeError("Input must be filename string or configuration dictionary.")
 
-    if config_file is None:
-        raise RuntimeError("No configure file provided")
 
-    if chosen_model is None:
-        raise RuntimeError("No model type provided")
+@run_prediction.register
```
allaffa, line 28: Why do we need this?

hydragnn/run_training.py
```diff
@@ -24,7 +25,23 @@
 from hydragnn.train.train_validate_test import train_validate_test
 
 
-def run_training(config_file="./examples/configuration.json"):
+@singledispatch
```
allaffa, line 28: Same question. I don't understand what was the problem before.

hydragnn/unit_tests/test_graphs.py
```diff
@@ -70,20 +71,20 @@ def pytest_train_model(model_type, ci_input, overwrite_data=False):
...
-    with open(tmp_file, "w") as f:
-        json.dump(config, f)
-
-    hydragnn.run_training(tmp_file)
+    # Since the config file uses PNA already, test the file overload here.
+    # All the other models need to use the locally modified dictionary.
+    if model_type == "PNA":
+        hydragnn.run_training(config_file)
+    else:
+        hydragnn.run_training(config)
```
allaffa, line 79: Very nice!

allaffa (approved):

hydragnn/run_prediction.py
streeve: (Posting from our private chat for posterity) We want to run the same function with a string or a dict, but Python can't figure out which function to use on its own, so we need all this extra syntax

Issue #2, opened 2021-10-19
Add edge features and invariances


streeve:
- [x] Edge lengths
- [x] Angles
- [x] Unit test translational invariance
- [x] Unit test rotational invariance

PR #11, opened 2021-10-20
Add edge lengths

streeve:
(From GCNN:)
With @pzhanggit 's help I have found:
- Only PNA and CGConv support
- GIN has an additional version GINE for edges
~~PNA currently breaks because we don't define the `edge_encoder`~~
~~GINE currently breaks because dimensions that don't agree (but I think actually do based on what I've checked)~~

Only implemented for PNA and CG (GINE does not have robust enough support to justify adding). New test only calls for these two models with edge lengths

Part of #2

### DIFF
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -194,7 +194,11 @@ class Base(Module):
         )
         ### encoder part ####
         for conv, batch_norm in zip(self.convs, self.batch_norms):
-            if data.edge_attr is not None:
+            if (data.edge_attr is not None) and (
+                hasattr(self, "edge_dim")
+                and self.edge_dim is not None
+                and self.edge_dim > 0
+            ):
                 c = conv(x=x, edge_index=edge_index, edge_attr=data.edge_attr)
             else:
                 c = conv(x=x, edge_index=edge_index)
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -206,12 +206,14 @@ class Base(Module):
             use_edge_attr = True
 
         ### encoder part ####
-        for conv, batch_norm in zip(self.convs, self.batch_norms):
-            if use_edge_attr:
+        if use_edge_attr:
+            for conv, batch_norm in zip(self.convs, self.batch_norms):
                 c = conv(x=x, edge_index=edge_index, edge_attr=data.edge_attr)
-            else:
+                x = F.relu(batch_norm(c))
+        else:
+            for conv, batch_norm in zip(self.convs, self.batch_norms):
                 c = conv(x=x, edge_index=edge_index)
-            x = F.relu(batch_norm(c))
+                x = F.relu(batch_norm(c))
 
         #### multi-head decoder part####
         # shared dense layers for graph level output
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -194,7 +194,12 @@ class Base(Module):
         )
         ### encoder part ####
         for conv, batch_norm in zip(self.convs, self.batch_norms):
-            x = F.relu(batch_norm(conv(x=x, edge_index=edge_index)))
+            if data.edge_attr is not None:
+                c = conv(x=x, edge_index=edge_index, edge_attr=data.edge_attr)
+            else:
+                c = conv(x=x, edge_index=edge_index)
+            x = F.relu(batch_norm(c))
+
         #### multi-head decoder part####
         # shared dense layers for graph level output
         x_graph = global_mean_pool(x, batch)
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -72,6 +72,15 @@ class Base(Module):
             weightabssum = sum(abs(number) for number in self.loss_weights)
             self.loss_weights = [iw / weightabssum for iw in self.loss_weights]
 
+        # Condition to pass edge_attr through forward propagation.
+        self.use_edge_attr = False
+        if (
+            hasattr(self, "edge_dim")
+            and self.edge_dim is not None
+            and self.edge_dim > 0
+        ):
+            self.use_edge_attr = True
+
         self._init_conv()
         self._init_node_conv()
         self._multihead()
@@ -192,13 +201,13 @@ class Base(Module):
             data.edge_index,
             data.batch,
         )
+        use_edge_attr = False
+        if (data.edge_attr is not None) and (self.use_edge_attr):
+            use_edge_attr = True
+
         ### encoder part ####
         for conv, batch_norm in zip(self.convs, self.batch_norms):
-            if (data.edge_attr is not None) and (
-                hasattr(self, "edge_dim")
-                and self.edge_dim is not None
-                and self.edge_dim > 0
-            ):
+            if use_edge_attr:
                 c = conv(x=x, edge_index=edge_index, edge_attr=data.edge_attr)
             else:
                 c = conv(x=x, edge_index=edge_index)
--- a/hydragnn/preprocess/serialized_dataset_loader.py
+++ b/hydragnn/preprocess/serialized_dataset_loader.py
@@ -69,9 +69,19 @@ class SerializedDataLoader:
         )
         compute_edge_lengths = Distance(norm=False, cat=True)
 
+        dataset[:] = [compute_edges(data) for data in dataset]
+        dataset[:] = [compute_edge_lengths(data) for data in dataset]
+
+        max_edge_length = torch.Tensor([float("-inf")])
+
         for data in dataset:
-            data = compute_edges(data)
-            data = compute_edge_lengths(data)
+            max_edge_length = torch.max(max_edge_length, torch.max(data.edge_attr))
+
+        # Normalization of the edges
+        for data in dataset:
+            data.edge_attr = data.edge_attr / max_edge_length
+
+        for data in dataset:
             self.__update_predicted_values(
                 config["Variables_of_interest"]["type"],
                 config["Variables_of_interest"]["output_index"],
--- a/hydragnn/preprocess/serialized_dataset_loader.py
+++ b/hydragnn/preprocess/serialized_dataset_loader.py
@@ -15,7 +15,7 @@ from sklearn.model_selection import StratifiedShuffleSplit
 
 import torch
 from torch_geometric.data import Data
-from torch_geometric.transforms import RadiusGraph
+from torch_geometric.transforms import RadiusGraph, Distance
 
 from .dataset_descriptors import AtomFeatures
 from hydragnn.utils.print_utils import print_distributed, iterate_tqdm
@@ -67,9 +67,11 @@ class SerializedDataLoader:
             loop=False,
             max_num_neighbors=config["Architecture"]["max_neighbours"],
         )
+        compute_edge_lengths = Distance(norm=True, cat=True)
 
         for data in dataset:
             data = compute_edges(data)
+            data = compute_edge_lengths(data)
             self.__update_predicted_values(
                 config["Variables_of_interest"]["type"],
                 config["Variables_of_interest"]["output_index"],
--- a/hydragnn/preprocess/serialized_dataset_loader.py
+++ b/hydragnn/preprocess/serialized_dataset_loader.py
@@ -67,7 +67,7 @@ class SerializedDataLoader:
             loop=False,
             max_num_neighbors=config["Architecture"]["max_neighbours"],
         )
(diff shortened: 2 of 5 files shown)


streeve:
Waiting for review on this one

allaffa:
@streeve 
Reviewing this will be my task on Jira this week :)

hydragnn/unit_tests/test_graphs.py
```diff
@@ -33,6 +31,10 @@ def pytest_train_model(model_type, ci_input, overwrite_data=False):
         config = json.load(f)
     config["NeuralNetwork"]["Architecture"]["model_type"] = model_type
 
+    # Only run with edge lengths for models that support them.
+    if use_lengths:
+        config["NeuralNetwork"]["Architecture"]["edge_features"] = ["lengths"]
```
streeve, line 36: @allaffa I add the edge feature on the fly here

hydragnn/unit_tests/test_graphs.py
```diff
@@ -73,7 +75,7 @@ def pytest_train_model(model_type, ci_input, overwrite_data=False):
 
     # Since the config file uses PNA already, test the file overload here.
     # All the other models need to use the locally modified dictionary.
-    if model_type == "PNA":
+    if model_type == "PNA" and not use_lengths:
```
pzhanggit, line 78: The `if ` condition here is not needed.

streeve: But if we modify the config we can't use the original file

pzhanggit: We can use the config dictionary, right?

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -66,15 +66,22 @@ def load_serialized_data(self, dataset_path: str, config):
             _ = pickle.load(f)
             dataset = pickle.load(f)
 
-        edge_index, edge_distances = self.__compute_edges(
+        edge_index, edge_length = self.__compute_edges(
```
pzhanggit, line 69: @allaffa  I think you are right about something wrong with the normalization. 
@streeve  Could you check line 196 inside self.__compute_edges(...)
`   # Normalize the lengths using min-max normalization
        edge_lengths = (edge_lengths - min(edge_lengths)) / (
            max(edge_lengths) - min(edge_lengths)
        )`. Currently, it uses min/max edge_length in each graph/data sample for normalization, which varies in dataset.

hydragnn/unit_tests/test_graphs.py
streeve: Then we won't test the file interface and miss some bug

pzhanggit: Aha, I see what you mean. It is for testing the file overload. Thanks

hydragnn/preprocess/serialized_dataset_loader.py
allaffa: Yep. the normalization should use `min` and `max` that span the entire dataset.

pzhanggit: Yeah, it is probably okay for now, given all the samples in the dataset have the same node position, but need to be improved.

allaffa:
@streeve 
The configuration file `examples/configuraton.json` still causes an error when PNA is used. 
PNA currently breaks because we don't define the edge_encoder. 
I saw that this error was crossed out, so I assumed the issue was solved. But I still get this error when I run the version of the code on this branch.

allaffa:
@streeve 

The constructor of the `PNA` model that overwrites the constructor of the `Base` model in `PNAStack.py` does not have edge attributes as input parameters. I suspect that the edge length is never used. Am I wrong?

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -84,6 +97,15 @@ def load_serialized_data(self, dataset_path: str, config):
...
+        # Normalize the lengths using min-max normalization
+        for data in dataset:
+            if (
+                "edge_features" in config["Architecture"]
+                and config["Architecture"]["edge_features"]
+            ):
+                #data.edge_attr = (data.edge_attr - self.min_edge_lengh_global)/( self.max_edge_lengh_global - self.min_edge_lengh_global )
+                data.edge_attr = (data.edge_attr)/( self.max_edge_lengh_global )
+
         if "subsample_percentage" in config["Variables_of_interest"].keys():
```
allaffa, line 109: @streeve 
After PR#30 has been merged, I will rebase this one and use `torch_geometric.tramsforms.distance` to compute the edge length. This transformation also provide a Boolean variable to automatically normalize all the edge lengths ion the range (0,1) if it is set to True. 
https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/transforms/distance.html#Distance

hydragnn/models/Base.py
```diff
@@ -194,7 +195,16 @@ def forward(self, data):
         )
         ### encoder part ####
         for conv, batch_norm in zip(self.convs, self.batch_norms):
-            x = F.relu(batch_norm(conv(x=x, edge_index=edge_index)))
+            if (data.edge_attr is not None) and (
```
pzhanggit, line 198: Maybe move this outside of the for loop? We only need to assess data.edge_attr once.

hydragnn/models/Base.py
```diff
@@ -195,7 +195,11 @@ def forward(self, data):
         )
         ### encoder part ####
         for conv, batch_norm in zip(self.convs, self.batch_norms):
-            if (data.edge_attr is not None) and (self.edge_dim is not None):
+            if (data.edge_attr is not None) and (
+                hasattr(self, "edge_dim")
```
streeve, line 199: @allaffa is this for the edge case where the data has `edge_attr`, but the model doesn't support it?

hydragnn/models/Base.py
streeve: Done in the last commit

streeve:
Cleaned up and rebased on main - thanks for the help @allaffa. This is ready for another review

allaffa:
> Cleaned up and rebased on main - thanks for the help @allaffa. This is ready for another review @pzhanggit

Your commit history is so clean and every commit is so clearly mapped into changes made to the code...

hydragnn/models/Base.py
allaffa: Correct.

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -67,9 +67,11 @@ def load_serialized_data(self, dataset_path: str, config):
             loop=False,
             max_num_neighbors=config["Architecture"]["max_neighbours"],
         )
+        compute_edge_lengths = Distance(norm=True, cat=True)
```
pzhanggit, line 70: Setting `norm=True` would normalize the distance with the maximum distance in each data sample. Is this by design?

allaffa: I was thinking Both this right now. I want to set it to `False`

hydragnn/models/Base.py
```diff
@@ -192,9 +202,18 @@ def forward(self, data):
...
             data.batch,
         )
+        use_edge_attr = False
+        if (data.edge_attr is not None) and (self.use_edge_attr):
+            use_edge_attr = True
+
         ### encoder part ####
         for conv, batch_norm in zip(self.convs, self.batch_norms):
-            x = F.relu(batch_norm(conv(x=x, edge_index=edge_index)))
+            if use_edge_attr:
```
pzhanggit, line 211: I was thinking a structure like,
```
if use_edge_attr:
       for conv, batch_norm in zip(self.convs, self.batch_norms):
             c = conv(x=x, edge_index=edge_index, edge_attr=data.edge_attr)
             x = F.relu(batch_norm(c))
else:
       for conv, batch_norm in zip(self.convs, self.batch_norms):
             c = conv(x=x, edge_index=edge_index)
             x = F.relu(batch_norm(c))
```
Not sure how much difference it could make though.

allaffa: The last commit from Sam does that. The relu call is out of the if-else condition though. 

if use_edge_attr:
       for conv, batch_norm in zip(self.convs, self.batch_norms):
             c = conv(x=x, edge_index=edge_index, edge_attr=data.edge_attr)
else:
       for conv, batch_norm in zip(self.convs, self.batch_norms):
             c = conv(x=x, edge_index=edge_index)

 x = F.relu(batch_norm(c))

pzhanggit: > 

Not really. For the current code, it still needs to run if-else for every `conv` inside for loop. [Btw, `relu` should be inside for loop in your comment. ]

allaffa: @pzhanggit 
committed new changes to follow your suggestions. Take a look and let me know. I am debating now whether to NOT perform the normalization on edge_length at all or if we should perform it over the entire dataset

pzhanggit (changes requested):
It looks good to me. Could you run one case with edge feature on DFT dataset? Wondering how much accuracy improvement it can get.

hydragnn/preprocess/serialized_dataset_loader.py
pzhanggit: We probably still need normalization (otherwise edge feature likely will dominate everything,  given that all other features are normalized). Need to figure out a proper way to do the normalization.

hydragnn/models/Base.py
pzhanggit: Lol, see my comment on normalization above. In short, I think we should have normalization but need do it properly.

hydragnn/models/Base.py
```diff
@@ -15,6 +15,7 @@
 from torch_geometric.nn import global_mean_pool, BatchNorm
 from torch.nn import GaussianNLLLoss
 import sys
+import inspect
```
streeve, line 18: Remove?

hydragnn/preprocess/serialized_dataset_loader.py
allaffa: Done.

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -69,9 +69,21 @@ def load_serialized_data(self, dataset_path: str, config):
...
         compute_edge_lengths = Distance(norm=False, cat=True)
 
+        dataset[:]=[compute_edges(data) for data in dataset]
+        dataset[:]=[compute_edge_lengths(data) for data in dataset]
+
+        min_edge_length = torch.Tensor([float('inf')])
+        max_edge_length = torch.Tensor([float('-inf')])
+
+        for data in dataset:
+            min_edge_length = torch.min(min_edge_length, torch.min(data.edge_attr))
```
streeve, line 79: Comment out since we're not using?

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -69,9 +69,21 @@ def load_serialized_data(self, dataset_path: str, config):
         )
         compute_edge_lengths = Distance(norm=False, cat=True)
 
+        dataset[:]=[compute_edges(data) for data in dataset]
```
streeve: This is nice

allaffa: @streeve Done.

hydragnn/preprocess/serialized_dataset_loader.py
allaffa: @streeve Done.

allaffa:
> It looks good to me. Could you run one case with edge feature on DFT dataset? Wondering how much accuracy improvement it can get.

@pzhanggit no improvement seen. I will open a separate PR to inver the edge length as edge attributes. Will keep you posted.

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -67,9 +67,22 @@ def load_serialized_data(self, dataset_path: str, config):
             loop=False,
             max_num_neighbors=config["Architecture"]["max_neighbours"],
         )
+        compute_edge_lengths = Distance(norm=False, cat=True)
+
+        dataset[:] = [compute_edges(data) for data in dataset]
+        dataset[:] = [compute_edge_lengths(data) for data in dataset]
+
+        min_edge_length = torch.Tensor([float("inf")])
```
pzhanggit, line 75: Is it used anywhere?

streeve: Thanks, fixed

pzhanggit (approved):

PR #12, opened 2021-10-20
Multi graph size 

pzhanggit:


hydragnn/models/Base.py
```diff
@@ -96,11 +71,71 @@ def _multihead(
...
+        for _ in range(self.num_conv_layers - 1):
+            conv = self.get_conv(self.hidden_dim, self.hidden_dim)
+            self.convs.append(conv)
+            self.batch_norms.append(BatchNorm(self.hidden_dim))
+
+    def _init_node_conv(self):
+        # *******convolutional layers for node level predictions*******#
+        # two ways to implement node features from here:
+        # 1. one graph for all node features
+        # 2. one graph for one node features (currently implemented)
```
streeve, line 91: Aren't both implemented?

streeve (approved):
Looks good to me - should get Max's approval as well though

hydragnn/models/Base.py
pzhanggit: No, only the second one.

allaffa (changes requested):
@pzhanggit 
The `configuration.json` file still has the number of nodes as input. If the PR is able to handle graphs of multiple size, this input should be removed and not used anymore.

pzhanggit:
> @pzhanggit The `configuration.json` file still has the number of nodes as input. If the PR is able to handle graphs of multiple size, this input should be removed and not used anymore.



Good point. Will remove it

pzhanggit:
> > @pzhanggit The `configuration.json` file still has the number of nodes as input. If the PR is able to handle graphs of multiple size, this input should be removed and not used anymore.
> 
> Good point. Will remove it

See b7b4644ab18a071e6951886fa717911b5d329ba0.

allaffa (approved):

PR #15, opened 2021-10-26
Timers

allaffa:


hydragnn/run_training.py
```diff
@@ -242,3 +243,5 @@ def _(config: dict):
             model.state_dict(),
             "./logs/" + model_with_config_name + "/" + model_with_config_name + ".pk",
         )
+
+    print_timers(verbosity)
```
streeve, line 247: We should print in `run_prediction` as well

streeve (changes requested):
There are some unresolved comments plus one more new one

hydragnn/run_prediction.py
```diff
@@ -126,3 +127,4 @@ def run_prediction(config: dict):
     ) = test(test_loader, model, config["Verbosity"]["level"])
 
     return error, error_sumofnodes_task, error_rmse_task, true_values, predicted_values
+    print_timers(verbosity)
```
streeve, line 130: Before return :)

hydragnn/utils/time_utils.py
```diff
@@ -0,0 +1,123 @@
...
+        )
+    ]
+
+    # The statistics are the result of global collective operations, so we only print them once
+    if verbosity >= 1:
+        if world_size > 1:
+            print_distributed(1, "Minimum timers: ")
+            [
+                print_distributed(1, key, " : ", round(value, 2))
+                for key, value in Timer.timers_min.items()
```
streeve, line 110: Add the same sorting as individual timers

hydragnn/run_prediction.py
allaffa: LOL. Thanks :)

streeve (approved):

PR #22, opened 2021-11-05
Add denormalization option in run_predictions

allaffa:


allaffa:
@seheracer 
I added the de-normalization option in `run_preduct.py`

hydragnn/run_prediction.py
```diff
@@ -125,4 +125,10 @@ def run_prediction(config: dict):
         predicted_values,
     ) = test(test_loader, model, config["Verbosity"]["level"])
 
+    ##output predictions with unit/not normalized
+    if config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"] == "True":
```
streeve, line 129: @pzhanggit we'll need to change this in your branch once this is merged to match your conversion to bool

streeve (approved):

streeve:
@allaffa conflicts are holding up the CI

hydragnn/run_prediction.py
pzhanggit: Agreed. Also, I am thinking if we need to run a thorough code check before and after modularization/organization to avoid problems like this one.

pzhanggit (approved):

PR #23, opened 2021-11-09
Update unit test imports

streeve:
Should not have imported module inside the module


streeve:
Optional merge - I don't think this is necessary, but is more consistent with the rest of the library

PR #25, opened 2021-11-15
setup file added

allaffa:
Add `setup.py`. Add installation files to `.gitignore`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,0 +1,22 @@
+import os
+from setuptools import setup, find_packages
+
+# Note: setup() has access to cmd arguments of the setup.py script via sys.argv
+
+# Utility function to read the README file.
+def read(fname):
+    return open(os.path.join(os.path.dirname(__file__), fname)).read()
+
+
+setup(
+    name="HydraGNN",
+    version="1.0",
+    package_dir={"hydragnn": "hydragnn"},
+    packages=find_packages(),
+    description="Distributed PyTorch implementation of multi-headed graph convolutional neural networks",
+    license="BSD-3",
+    long_description=read("README.md"),
+    url="https://github.com/ORNL/HydraGNN",
+    author="Massimiliano Lupo Pasini, Samuel Temple Reeve, Pei Zhang, Jong Youl Choi",
+    author_email="lupopasinim@ornl.gov",
+)
--- a/setup.py
+++ b/setup.py
@@ -8,11 +8,25 @@ def read(fname):
     return open(os.path.join(os.path.dirname(__file__), fname)).read()
 
 
+install_requires = [
+    "pickle5",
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -0,0 +1,12 @@
+from distutils.core import setup
+
+# Note: setup() has access to cmd arguments of the setup.py script via sys.argv
+
+setup(
+    name="HydraGNN",
+    version="1.0",
```
streeve, line 7: I would start with 0.1-dev or 0.1

setup.py
```diff
@@ -0,0 +1,12 @@
...
+    package_dir={"hydragnn": "hydragnn"},
+    packages=["hydragnn"],
+    author="Massimiliano Lupo Pasini, Samuel Temple Reeve, Pei Zhang, Jong Youl Choi",
```
streeve, line 10: Marko?

streeve (changes requested):

streeve:
What you started with is probably sufficient, but we may as well add other easy categories (e.g. license): https://pythonhosted.org/an_example_pypi_project/setuptools.html 

Do you want to replace `requirements.txt` or keep both?

setup.py
allaffa: He is nor an author of the repository

allaffa:
> What you started with is probably sufficient, but we may as well add other easy categories (e.g. license): https://pythonhosted.org/an_example_pypi_project/setuptools.html
> 
> Do you want to replace `requirements.txt` or keep both?

@streeve I would like to keep both

setup.py
streeve: After discussing we can leave as is and plan to release after current PRs are mereged

setup.py
streeve: Not actively, but his contributions are still in the repo

setup.py
```diff
@@ -0,0 +1,16 @@
...
+# Note: setup() has access to cmd arguments of the setup.py script via sys.argv
+
+setup(
+    name="HydraGNN",
+    version="1.0",
+    package_dir={"hydragnn": "hydragnn"},
+    packages=["hydragnn"],
+    description="Distributed PyTorch implementation of multi-headed graph convolutional neural networks",
+    license="BSD-3",
+    long_description=read("README"),
```
streeve, line 12: Need to add the `read` helper function

streeve (changes requested):

setup.py
allaffa: The list of people added here is going to reflect the name of people listed as authors in the technical transfer.

pzhanggit (approved):

streeve (approved):

PR #31, opened 2021-11-16
Add edge length as input feature

allaffa:


hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -67,9 +67,11 @@ def load_serialized_data(self, dataset_path: str, config):
             loop=True,
             max_num_neighbors=config["Architecture"]["max_neighbours"],
         )
+        compute_edge_lengths = Distance(norm=True, cat=True)
```
streeve, line 70: It would be nice if there was a way to fuse these operations together, but I don't see anything obvious

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -67,9 +67,11 @@ def load_serialized_data(self, dataset_path: str, config):
...
 
         for data in dataset:
             data = compute_edges(data)
+            data = compute_edge_lengths(data)
```
streeve, line 74: Shouldn't we only store these if we need them?

allaffa: I don't know how to make this PR as "draft". I don't find the button anymore

PR #32, opened 2021-11-18
Improve dataset loading

pzhanggit:
1. fixed a bug when loading datasets separately for training/test/validate
2. remove the need that dataset has to be one of the prescribed in dataset_descriptors.py
3. fix a bug related to  https://github.com/ORNL/HydraGNN/pull/22

### DIFF
--- a/hydragnn/preprocess/load_data.py
+++ b/hydragnn/preprocess/load_data.py
@@ -23,66 +23,22 @@ except:
 
 from hydragnn.preprocess.serialized_dataset_loader import SerializedDataLoader
 from hydragnn.preprocess.raw_dataset_loader import RawDataLoader
-from hydragnn.preprocess.dataset_descriptors import Dataset
 from hydragnn.utils.distributed import get_comm_size_and_rank
 from hydragnn.utils.time_utils import Timer
 
 
 def dataset_loading_and_splitting(
     config: {},
-    chosen_dataset_option: Dataset,
+    chosen_dataset_option,
 ):
 
-    if chosen_dataset_option in [item.value for item in Dataset]:
-        dataset_chosen, dataset_names = load_data(chosen_dataset_option, config)
-        return split_dataset(
-            dataset_list=dataset_chosen,
-            dataset_names=dataset_names,
-            batch_size=config["NeuralNetwork"]["Training"]["batch_size"],
-            perc_train=config["NeuralNetwork"]["Training"]["perc_train"],
-        )
-    else:
-        # FIXME, should re-normalize mixed datasets based on joint min_max
-        raise ValueError(
-            "Chosen dataset option not yet supported", chosen_dataset_option
-        )
-        dataset_CuAu = load_data(Dataset.CuAu.value, config)
-        dataset_FePt = load_data(Dataset.FePt.value, config)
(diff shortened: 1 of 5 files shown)


allaffa (approved):

hydragnn/preprocess/load_data.py
```diff
@@ -23,65 +23,22 @@
 
 from hydragnn.preprocess.serialized_dataset_loader import SerializedDataLoader
 from hydragnn.preprocess.raw_dataset_loader import RawDataLoader
-from hydragnn.preprocess.dataset_descriptors import Dataset
```
streeve, line 26: Let's remove if no longer using

streeve (commented):
@pzhanggit you're not allowed to title "fix bug" anymore 👎

hydragnn/preprocess/load_data.py
pzhanggit: Removed in 5a35268f9739410d224d78e4d542b966b4870161

hydragnn/run_training.py
```diff
@@ -77,7 +77,14 @@ def _(config: dict):
         and config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"]
         == "True"
     ):
-        dataset_path = f"{os.environ['SERIALIZED_DATA_PATH']}/serialized_dataset/{config['Dataset']['name']}.pkl"
+        for dataset_name, _ in config["Dataset"]["path"]["raw"].items():
```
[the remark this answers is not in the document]
pzhanggit: Done in 83fa1d1e03601f18c5b036a410d8cef71a83cd6f

hydragnn/run_prediction.py
```diff
@@ -130,7 +131,9 @@ def _(config: dict):
     ##output predictions with unit/not normalized
     if config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"] == "True":
         true_values, predicted_values = output_denormalize(
-            config["Variables_of_interest"]["y_minmax"], true_values, predicted_values
+            config["NeuralNetwork"]["Variables_of_interest"]["y_minmax"],
```
streeve, line 134: Did this not fail because we don't test it?

hydragnn/run_training.py
```diff
@@ -77,7 +77,11 @@ def _(config: dict):
         and config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"]
         == "True"
     ):
-        dataset_path = f"{os.environ['SERIALIZED_DATA_PATH']}/serialized_dataset/{config['Dataset']['name']}.pkl"
+        if "total" in config["Dataset"]["path"]["raw"].items():
+            dataset_path = f"{os.environ['SERIALIZED_DATA_PATH']}/serialized_dataset/{config['Dataset']['name']}.pkl"
+        else:
+            ###used for min/max values loading below
```
streeve, line 83: add "and so we only need to read one file"

hydragnn/run_prediction.py
pzhanggit: It was added recently and is not tested in current unit test. For other runs, I do not know.

hydragnn/run_training.py
pzhanggit: I think this is good enough.

PR #34, opened 2021-11-19
Move tests to top level

streeve:
Now a separate module from `hydragnn`


allaffa:
@streeve 
run_training.py and run_predictions.py are currently still under the hydragnn module. 
should we move them in the examples?

streeve:
This PR only touches tests - I wanted to do that separately because the tests depend on those files right now

allaffa (approved):

Issue #5, opened 2021-10-19
Test with external datasets


streeve:
- [x] torch_geometric.datasets
  - QM9
  - MD17
- [ ] Materials Project
- [ ] OQMD
- [ ] Matbench
etc.

allaffa:
Completed.

PR #35, opened 2021-11-19
Add torch_geometric datasets

streeve:
- QM9 molecular dataset
- MD17 molecular trajectory dataset

Part of #5 

Need to install to run: `python setup.py develop` or `python setup.py install`
OR add to the python path as in the CI

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -47,6 +47,8 @@ jobs:
           git diff --exit-code
       - name: Run pytest
         run: |
+          # FIXME: install
+          export PYTHONPATH=$PYTHONPATH:~/HydraGNN
           pip list
           python -m pytest -rfEP -W error -W ignore::DeprecationWarning
           mpirun -n 2 python -m pytest -rfEP -W error -W ignore::DeprecationWarning --with-mpi
--- a/examples/md17/md17.py
+++ b/examples/md17/md17.py
@@ -1,0 +1,87 @@
+import os, json
+
+import torch
+import torch_geometric
+
+import hydragnn
+
+# Update each sample prior to loading.
+def md17_pre_transform(data):
+    # Set descriptor as element type.
+    data.x = data.z.float()
+    # Only predict energy (index 0 of 2 properties) for this run.
+    data.y = data.energy
+    hydragnn.preprocess.update_predicted_values(
+        var_config["type"],
+        var_config["output_index"],
+        data,
+    )
+    data = compute_edges(data)
+
+    return data
+
+
+# Randomly select ~1000 samples
+def md17_pre_filter(data):
+    return torch.rand(1) < 0.25
+
+
+# Set this path for output.
+try:
+    os.environ["SERIALIZED_DATA_PATH"]
+except:
+    os.environ["SERIALIZED_DATA_PATH"] = os.getcwd()
+
+# Configurable run choices (JSON file that accompanies this example script).
+filename = os.path.join(os.path.dirname(__file__), "md17.json")
+with open(filename, "r") as f:
+    config = json.load(f)
+verbosity = config["Verbosity"]["level"]
+arch_config = config["NeuralNetwork"]["Architecture"]
+var_config = config["NeuralNetwork"]["Variables_of_interest"]
+
+# Set the details of outputs of interest.
+arch_config["output_dim"] = [0]
+arch_config["output_type"] = ["graph"]
+
+# Always initialize for multi-rank training.
+world_size, world_rank = hydragnn.utils.distributed.setup_ddp()
+
+# Use built-in torch_geometric dataset.
+compute_edges = hydragnn.preprocess.get_radius_graph(arch_config)
+dataset = torch_geometric.datasets.MD17(
+    root="dataset/md17",
+    name="uracil",
+    pre_transform=md17_pre_transform,
+    pre_filter=md17_pre_filter,
+)
+train_loader = torch_geometric.loader.DataLoader(dataset, batch_size=32, shuffle=False)
+
+# Add model changes to config.
+config["NeuralNetwork"]["Architecture"] = arch_config
+config["NeuralNetwork"]["Variables_of_interest"] = var_config
+
+input_dim = len(config["NeuralNetwork"]["Variables_of_interest"]["input_node_features"])
+model = hydragnn.models.create(
+    model_type=arch_config["model_type"],
+    input_dim=input_dim,
+    dataset=dataset,
+    config=arch_config,
+    verbosity_level=verbosity,
+)
+
+learning_rate = config["NeuralNetwork"]["Training"]["learning_rate"]
+optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate)
+scheduler = torch.optim.lr_scheduler.ReduceLROnPlateau(
+    optimizer, mode="min", factor=0.5, patience=5, min_lr=0.00001
+)
+
+# Run training with the given model and md17 dataset.
+num_epoch = config["NeuralNetwork"]["Training"]["num_epoch"]
+for epoch in range(0, num_epoch):
+    train_rmse, _, _ = hydragnn.train.train(train_loader, model, optimizer, verbosity)
+    scheduler.step(train_rmse)
+
+    hydragnn.utils.print_distributed(
+        verbosity, f"Epoch: {epoch:02d}, Train RMSE: {train_rmse:.8f}"
+    )
--- a/examples/md17/md17.py
+++ b/examples/md17/md17.py
@@ -3,12 +3,18 @@ import os, json
 import torch
 import torch_geometric
 
+# deprecated in torch_geometric 2.0
+try:
+    from torch_geometric.loader import DataLoader
+except:
+    from torch_geometric.data import DataLoader
+
 import hydragnn
 
 # Update each sample prior to loading.
 def md17_pre_transform(data):
     # Set descriptor as element type.
-    data.x = data.z.float()
+    data.x = data.z.float().view(-1, 1)
     # Only predict energy (index 0 of 2 properties) for this run.
     data.y = data.energy
     hydragnn.preprocess.update_predicted_values(
@@ -17,8 +23,8 @@ def md17_pre_transform(data):
         data,
     )
     data = compute_edges(data)
-
-    return data
+    device = hydragnn.utils.get_device()
+    return data.to(device)
 
 
 # Randomly select ~1000 samples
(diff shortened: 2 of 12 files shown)


allaffa:
@streeve 

Thank you for doing this. 
What I have in mind is to provide the users with examples of simple, intermediate, and advanced difficulty. 

**Simple**: use datasets directly provided in` torch.geometric` and explicitly build the NN without configure files and without using `hydrogen/models/create.py `
**Intermediate**: use datasets directly provided in `torch.geometric` and explicitly build the NN using configure files that are passed to `hydrogen/models/create.py `
**Advanced**: use personalized datasets directly provided by raw data files and explicitly build the NN using configure files that are passed to `hydrogen/models/create.py `

Some of the **simple** examples (those ones that can be run in a reasonably short amount of time) would also be good candidates to become unit tests. 

What you provide is a very good **intermediate**.

streeve:
The current issue on this PR is that the PyG QM9 dataset is immutable - we'll have to create a new object to add extra fields for multi-head

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -102,30 +98,6 @@ def __update_atom_features(self, atom_features: [AtomFeatures], data: Data):
         feature_indices = [i for i in atom_features]
         data.x = data.x[:, feature_indices]
 
-    def __update_predicted_values(self, type: list, index: list, data: Data):
```
allaffa, line 105: @streeve 
Why are you leaving `self.__update_atom_features` as a method of `SerializedDataLoader` and extracting `update_predicted_values` as an external standalone function?

allaffa (changes requested):

streeve:
Keeping this open since there are no differences in #42 after rebasing on main

hydragnn/preprocess/serialized_dataset_loader.py
streeve: I pulled that function out to use it in the new example preprocess functions - which was very easy because it didn't use any class data

streeve:
@pzhanggit this is up to date with main if you have time to take a look

examples/md17/md17.py
```diff
@@ -0,0 +1,86 @@
...
+# Configurable run choices (JSON file that accompanies this example script).
+filename = os.path.join(os.path.dirname(__file__), "md17.json")
+with open(filename, "r") as f:
+    config = json.load(f)
+verbosity = config["Verbosity"]["level"]
+arch_config = config["NeuralNetwork"]["Architecture"]
+var_config = config["NeuralNetwork"]["Variables_of_interest"]
+
+# Set the details of outputs of interest.
+arch_config["output_dim"] = [0]
```
pzhanggit, line 37: Why is output_dim equal to 0 not 1 for energy?

streeve:
> Are the codes ready to run?

They still fail on tensor mismatch, but I do not know what the issue is

pzhanggit:
> > Are the codes ready to run?
> 
> They still fail on tensor mismatch, but I do not know what the issue is
Error message?

streeve:
For tests to run we either need to update how we install or hack it in for now

examples/md17/md17.py
streeve: Seemed reasonable at the time

streeve:
@jychoi-hpc this seems to be failing in the CI because of the profiler; can you take a look?

jychoi-hpc:
Sure, I am looking at now.

jychoi-hpc:
I just made some changes to fix the error with profile:
* replaced "contextlib.nullcontext" with "Profiler()". Now profiler can act as a null.
* update on return values. Replaced `train_rmse, _, _ = train(...)` with `train_rmse, _  = train(...)` as `train()` returns only 2 values.

examples/md17/md17.py
```diff
@@ -0,0 +1,87 @@
...
+# Configurable run choices (JSON file that accompanies this example script).
+filename = os.path.join(os.path.dirname(__file__), "md17.json")
+with open(filename, "r") as f:
+    config = json.load(f)
+verbosity = config["Verbosity"]["level"]
+arch_config = config["NeuralNetwork"]["Architecture"]
+var_config = config["NeuralNetwork"]["Variables_of_interest"]
+
+# Set the details of outputs of interest.
+arch_config["output_dim"] = [1]
```
pzhanggit, line 44: Could we specify this in json input file?

streeve:
Ran full QM9 dataset for 5 epoch test; looks okay for a start, but needs some work

Screen Shot 2022-01-04 at 3 10 54 PM

examples/md17/md17.py
```diff
@@ -0,0 +1,109 @@
...
+    from torch_geometric.data import DataLoader
+
+import hydragnn
+
+# Update each sample prior to loading.
+def md17_pre_transform(data):
+    # Set descriptor as element type.
+    data.x = data.z.float().view(-1, 1)
+    # Only predict energy (index 0 of 2 properties) for this run.
+    data.y = data.energy / len(data.x)
```
allaffa, line 19: Why are you dividing the energy by the number of nodes in the graph?

examples/qm9/qm9.py
```diff
@@ -0,0 +1,105 @@
...
+    from torch_geometric.data import DataLoader
+
+import hydragnn
+
+# Update each sample prior to loading.
+def qm9_pre_transform(data):
+    # Set descriptor as element type.
+    data.x = data.z.float().view(-1, 1)
+    # Only predict free energy (index 10 of 19 properties) for this run.
+    data.y = data.y[:, 10] / len(data.x)
```
allaffa, line 19: Same question as for MD17. Why are you dividing by the size of the graph (i.e., number of nodes)?

allaffa (changes requested):

examples/md17/md17.py
streeve: This improves the prediction. It's the same unit conversion that Pei added for LSMS, in this case from eV to eV/atom

streeve:
Updated prediction scaling energy by number of atoms:
Screen Shot 2022-01-06 at 11 18 49 AM

examples/md17/md17.py
pzhanggit: Just curious, for variables in the two datasets, do they have some physical relationships similar to the one in https://github.com/ORNL/HydraGNN/pull/56?

pzhanggit:
> Updated prediction scaling energy by number of atoms: Screen Shot 2022-01-06 at 11 18 49 AM

Nice. It seems improved.

examples/md17/md17.py
```diff
@@ -0,0 +1,109 @@
...
+    root="dataset/md17",
+    name="uracil",
+    pre_transform=md17_pre_transform,
+    pre_filter=md17_pre_filter,
+)
+train, val, test = hydragnn.preprocess.split_dataset(
+    dataset, config["NeuralNetwork"]["Training"]["perc_train"]
+)
+bs = config["NeuralNetwork"]["Training"]["batch_size"]
+train_loader = DataLoader(train, batch_size=bs, shuffle=False)
```
pzhanggit, line 71: Is distributed used in the two examples?

examples/qm9/qm9.py
```diff
@@ -0,0 +1,105 @@
...
+filename = os.path.join(os.path.dirname(__file__), "qm9.json")
+with open(filename, "r") as f:
+    config = json.load(f)
+verbosity = config["Verbosity"]["level"]
+arch_config = config["NeuralNetwork"]["Architecture"]
+var_config = config["NeuralNetwork"]["Variables_of_interest"]
+
+# Set the details of outputs of interest.
+arch_config["output_dim"] = [1]
+arch_config["output_type"] = ["graph"]
```
pzhanggit, line 51: From json file?

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -191,3 +161,34 @@ def __stratified_sampling(self, dataset: [Data], subsample_percentage: float):
...
+    """Updates values of the structure we want to predict. Predicted value is represented by integer value.
+    Parameters
+    ----------
+    type: "graph" level or "node" level
+    index: index/location in data.y for graph level and in data.x for node level
+    data: Data
+        A Data object representing a structure that has atoms.
+    """
+    output_feature = []
+    data.y_loc = torch.zeros(1, len(type) + 1, dtype=torch.int64)
```
pzhanggit, line 184:
```suggestion
    data.y_loc = torch.zeros(1, len(type) + 1, dtype=torch.int64, device=data.y.device)
```

pzhanggit:
It looks alright to me. Only some minor comments

examples/md17/md17.py
streeve: There are a number of thermodynamic variables in the qm9 dataset - right now it's only using free energy. For any of the energetic values I think we will want to change units to energy/atom

examples/qm9/qm9.py
streeve: Oh, I should use `update_config_NN_outputs`, right? That's why it's not in any other config file. This is confusing as a user :)

hydragnn/preprocess/serialized_dataset_loader.py
streeve: I'll test on Summit and add any others I missed

examples/md17/md17.py
streeve: Good point - I was lazy and did this here instead of updating the library loading functions

streeve: Fixed

examples/qm9/qm9.py
streeve: Fixed with your update

hydragnn/preprocess/serialized_dataset_loader.py
streeve: Fixed

streeve:
@pzhanggit @allaffa can you do a final check?

.github/workflows/CI.yml
```diff
@@ -47,6 +47,8 @@ jobs:
           git diff --exit-code
       - name: Run pytest
         run: |
+          # FIXME: install
+          export PYTHONPATH=$PYTHONPATH:~/HydraGNN
```
pzhanggit, line 51: Thank you

hydragnn/run_training.py
```diff
@@ -25,6 +25,7 @@
     save_model,
     get_summary_writer,
     load_existing_model_config,
+    calculate_PNA_degree_config,
```
pzhanggit, line 28: No need to add since not used?

pzhanggit (approved):
Looks great. Thanks!

hydragnn/run_training.py
streeve: Good point - removed

streeve:
One more run - I tried increasing the hidden layers. Looks better for sure, but still a bias in the predictions at lower energy

free_energy

allaffa (approved):

PR #37, opened 2021-11-20
structure examples directory with multiple subdirectories

allaffa:
… example


tests/test_config.py
```diff
@@ -17,7 +17,7 @@
 @pytest.mark.mpi_skip()
 def pytest_config(config_file):
 
-    config_file = os.path.join("examples", config_file)
+    config_file = os.path.join("examples/dft_lsms", config_file)
```
streeve, line 20: Maybe just `lsms` for now? If we expand to examples outside DFT we can add another level to the hierarchy

allaffa: Done.

pzhanggit:
Need to update example.py file.

pzhanggit (approved):

pzhanggit:
> In the QM9 PR I've named `examples/qm9/qm9.json`. I would prefer `lsms/lsms.json` - I assume you disagree :)

Have to reply, lolllll

allaffa:
> In the QM9 PR I've named `examples/qm9/qm9.json`. I would prefer `lsms/lsms.json` - I assume you disagree :)

Done. I do not know the difference between agreeing and disagreeing anymore. :)

streeve (approved):

streeve (changes requested):
Approved, but can you squash these commits?

Issue #39, opened 2021-11-21
Load serialized *pkl files directly


pzhanggit:
Currently, dataset loading support raw data files, e.g., LSMS format. For every run, it would read the raw data files, convert them to serialized format and generate *pkl files. The process can be time consuming and even unnecessary sometimes. We should provide an option to load *pkl files directly.

PR #40, opened 2021-11-21
 Add loading pkl option

pzhanggit:
Closes https://github.com/ORNL/HydraGNN/issues/39

~~[Note: rebased on #38, should review/merge #38 first]~~

### DIFF
--- a/tests/test_graphs.py
+++ b/tests/test_graphs.py
@@ -18,7 +18,7 @@ import shutil
 import hydragnn, tests
 
 
-@pytest.mark.parametrize("model_type", ["GIN", "GAT", "MFC", "PNA", "CGCNN"])
+@pytest.mark.parametrize("model_type", ["PNA", "GIN", "GAT", "MFC", "CGCNN"])
 @pytest.mark.parametrize("ci_input", ["ci.json", "ci_multihead.json"])
 def pytest_train_model(model_type, ci_input, overwrite_data=False):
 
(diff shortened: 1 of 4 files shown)


allaffa (approved):

tests/test_graphs.py
```diff
@@ -35,41 +35,60 @@ def pytest_train_model(model_type, ci_input, overwrite_data=False):
...
+                     },
+            ...
+        """
+        ##check if serialized pickle files or folders for raw files provided
+        PKL_input = False
+        if list(config["Dataset"]["path"]["raw"].values())[0].endswith(".pkl"):
+            PKL_input = True
+        # only generate new datasets, if not pkl
+        if not PKL_input:
+            for dataset_name, data_path in config["Dataset"]["path"]["raw"].items():
```
streeve, line 56: Can we set up the tests here to use pickles if they exist instead? That will save us time and still test the data input for the first run

tests/test_graphs.py
```diff
@@ -35,41 +35,60 @@ def pytest_train_model(model_type, ci_input, overwrite_data=False):
...
                     )
+                elif dataset_name == "validate":
+                    num_samples = int(
+                        num_samples_tot
+                        * (1 - config["NeuralNetwork"]["Training"]["perc_train"])
+                        * 0.5
+                    )
+                if not os.listdir(data_path):
+                    num_nodes = config["Dataset"]["num_nodes"]
+                    if num_nodes == 4:
```
streeve, line 82: I think I'm going to fix this in a separate PR

tests/test_graphs.py
pzhanggit: Good point. For unit tests, we can change the config parameter after the first run so that the following would load pkl files.

tests/test_graphs.py
pzhanggit: I do not think I understand. Fix what?

streeve: It seems very strange to have a separate line for 4 atoms, but nothing else

pzhanggit: I see. Looking forward to the new PR.

tests/test_graphs.py
pzhanggit: See 0923e0505ae3167c98eddd40781dfff9c55f5ca0

tests/test_graphs.py
```diff
@@ -35,41 +35,73 @@ def pytest_train_model(model_type, ci_input, overwrite_data=False):
...
+            "path": {
+                "raw": {
+                    "train": "serialized_dataset/unit_test_singlehead_train.pkl",
+                    "test": "serialized_dataset/unit_test_singlehead_test.pkl",
+                    "validate": "serialized_dataset/unit_test_singlehead_validate.pkl"}
+                     },
+            ...
+        """
+        # starting from the second model (GIN is the first), use pkl files generated in the first model run by default
+        if model_type != "GIN":
```
streeve, line 51: I would check if the pickle files exist instead

streeve (approved):

allaffa (approved):

PR #43, opened 2021-11-28
output_denormalize treated as Boolean variable

allaffa:
the difference between true and True as string was causing the denormalization to never happen

### DIFF
--- a/hydragnn/run_prediction.py
+++ b/hydragnn/run_prediction.py
@@ -89,7 +89,7 @@ def _(config: dict):
     ) = test(test_loader, model, config["Verbosity"]["level"])
 
     ##output predictions with unit/not normalized
-    if config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"] == "True":
+    if bool(config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"]):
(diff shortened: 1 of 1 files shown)


hydragnn/run_prediction.py
```diff
@@ -89,7 +89,7 @@ def _(config: dict):
     ) = test(test_loader, model, config["Verbosity"]["level"])
 
     ##output predictions with unit/not normalized
-    if config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"] == "True":
+    if bool(config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"]):
```
streeve, line 92: Do we need to convert to bool here? I don't think we do anywhere else

pzhanggit: I also think we do not need to convert to bool here.

pzhanggit, line 92:
```suggestion
    if config["NeuralNetwork"]["Variables_of_interest"]["denormalize_output"]:
```

pzhanggit (changes requested):

pzhanggit (approved):

streeve (approved):

PR #44, opened 2021-11-29
add node type to configuration.json

pzhanggit:


pzhanggit:
> Can you add a simple test to stop this from happening again? Just running this file with one epoch in `test_config`

Any thoughts on how to deal with datasets for the suggested run?

allaffa (approved):

streeve:
> Any thoughts on how to deal with datasets for the suggested run?

Can you just create the minimum number of unit test samples for it to not crash?

pzhanggit:
> > Any thoughts on how to deal with datasets for the suggested run?
> 
> Can you just create the minimum number of unit test samples for it to not crash?

@streeve  Not sure I misunderstand you. Are you suggesting to upload DFT datasets for CI tests?

streeve:
> Not sure I misunderstand you. Are you suggesting to upload DFT datasets for CI tests?

No, just calling the `deterministic_graph_data` with a very small number of samples

pzhanggit:
> > Not sure I misunderstand you. Are you suggesting to upload DFT datasets for CI tests?
> 
> No, just calling the `deterministic_graph_data` with a very small number of samples

Sure, we can do that. The only downside is we need to manually change lots of config dicts due to different datasets. 
The other way (which I think also the better) maybe to simply compare dicts in /.examples/configuration.json and ./tests/inputs/ci_multihead.json and make sure all necessary dicts are there. What do you think?

streeve:
> simply compare dicts in /.examples/configuration.json and ./tests/inputs/ci_multihead.json and make sure all necessary dicts are there. What do you think?

Good point - your idea is better. Can you replace the current minimal `test_config` with that

streeve (approved):

streeve:
Will add test in a separate PR

PR #48, opened 2021-11-30
Create helper functions

streeve:
Moves most of `run_training` and `run_prediction` into separate functions - one step towards moving those to examples rather than in the library. 

Also adds some cases of functions with and without config inputs (always naming the function with a `_config` suffix), trying to make it easy to run either way.

File and function names are certainly up for debate


hydragnn/models/create.py
```diff
@@ -20,7 +20,7 @@
 from hydragnn.models.MFCStack import MFCStack
 from hydragnn.models.CGCNNStack import CGCNNStack
 
-from hydragnn.utils.device import get_device
+from hydragnn.utils.distributed import get_device
```
allaffa, line 23: Thank you for this! It makes much more sense here.

hydragnn/run_training.py
```diff
@@ -87,74 +84,33 @@ def _(config: dict):
...
-            model = torch.nn.parallel.DistributedDataParallel(model)
-        else:
-            model = torch.nn.parallel.DistributedDataParallel(
-                model, device_ids=[device]
-            )
-
-    optimizer = torch.optim.AdamW(
-        model.parameters(), lr=config["NeuralNetwork"]["Training"]["learning_rate"]
-    )
+    model = get_distributed_model(model, verbosity)
```
allaffa, line 92: Thank you for taking care of this. This was long due.

hydragnn/utils/model.py
```diff
@@ -0,0 +1,51 @@
+##############################################################################
```
allaffa, line 1: Maybe we can name this file "model_utils.py"? 
The word "model" is somewhat misleading, as people may expect to open the file and see the implementation of an actual deep learning model inside.

allaffa (changes requested):

hydragnn/utils/model.py
streeve: I was thinking about this too - only half the files in `utils/` have `_utils` in the name. I see your point, but a user should only need to use `from hydragnn.utils import x`. But I'm okay renaming all with or without `_utils`

streeve: @allaffa @pzhanggit strong feelings on this? I just think we should be consistent on the file names

hydragnn/utils/config_utils.py
```diff
@@ -42,32 +42,39 @@ def update_config_NN_outputs(config, graph_size_variable):
     return config
 
 
+def normalize_output_config(config):
+    var_config = config["NeuralNetwork"]["Variables_of_interest"]
```
pzhanggit, line 46: Neat!

allaffa (approved):

PR #50, opened 2021-12-01
Simplify model creation

streeve:
Similar to #48, updating model creation:

- adds version without config input
- separates config input from individual models
- makes inputs only needed for PNA optional

### DIFF
--- a/hydragnn/models/create.py
+++ b/hydragnn/models/create.py
@@ -12,7 +12,6 @@
 import os
 import torch
 from torch_geometric.data import Data
-from torch_geometric.utils import degree
 
 from hydragnn.models.GINStack import GINStack
 from hydragnn.models.PNAStack import PNAStack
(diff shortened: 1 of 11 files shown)


hydragnn/models/create.py
```diff
@@ -25,100 +25,127 @@
...
             input_dim=input_dim,
-            output_dim=config["output_dim"],
-            hidden_dim=config["hidden_dim"],
-            num_nodes=num_atoms,
-            num_conv_layers=config["num_conv_layers"],
-            output_type=config["output_type"],
-            config_heads=config["output_heads"],
-            loss_weights=config["task_weights"],
-        ).to(device)
+            output_dim=output_dim,
```
allaffa, line 81: @streeve 

This is beautiful, as it indeed makes the creation of the mode more practical when the configure file is not needed. For instance, when the data is immediately provided by torch.geometric

allaffa (approved):

hydragnn/models/create.py
```diff
@@ -25,100 +25,127 @@
...
-    verbosity_level: int,
+    output_dim: int,
+    hidden_dim: int,
+    num_conv_layers: int,
+    output_type: str,
+    output_heads: dict,
+    task_weights: list,
+    max_neighbours: int = 0,
+    num_nodes: int = None,
+    dataset: [Data] = None,
```
pzhanggit, line 67: Why do we need data in create models?

pzhanggit: Just realized PNA needs it for deg. A weak comment: is it better to just pass `deg` here instead of entire dataset?

pzhanggit (commented):
It looks good to me. Only have one minor comment.

allaffa (changes requested):
Just pass "deg" instead of dataset as suggested already by @pzhanggit

streeve:
Updated to separate `deg` for PNA, `max_neigh` for MFC, and `num_nodes` for MLP. I squash rebased because things got messy

hydragnn/utils/model.py
```diff
@@ -49,3 +51,14 @@ def load_existing_model(model, model_name, path="./logs/"):
     path_name = os.path.join(path, model_name, model_name + ".pk")
     state_dict = torch.load(path_name, map_location="cpu")
     model.load_state_dict(state_dict)
+
+
+def calculate_PNA_degree(dataset: [Data], max_neighbours):
+    num_nodes = dataset[0].num_nodes
```
pzhanggit, line 57: This line is not needed?

hydragnn/run_prediction.py
```diff
@@ -64,14 +65,20 @@ def _(config: dict):
...
-        dataset=train_loader.dataset,
+    config["NeuralNetwork"]["Architecture"]["input_dim"] = len(
+        config["NeuralNetwork"]["Variables_of_interest"]["input_node_features"]
+    )
+    max_neigh = config["NeuralNetwork"]["Architecture"]["max_neighbours"]
+    if config["NeuralNetwork"]["Architecture"]["model_type"] == "PNA":
+        deg = calculate_PNA_degree(train_loader.dataset, max_neigh)
+    else:
+        deg = None
+    model = create_model_config(
```
pzhanggit, line 76: A general comment that is not related to this PR specifically: why do we need to create the model in `run_prediction` (and then load model state )? Could we skip the model creation and just load a trained model directly?

pzhanggit (approved):
Thanks for the PR. It looks great now.

hydragnn/run_prediction.py
streeve: I think we could do that if we save the entire model instead of just the state_dict. Not sure that's worth it

hydragnn/utils/model.py
streeve: Fixed

hydragnn/run_prediction.py
pzhanggit: Good point, it seems the current way is better and more robust, https://pytorch.org/tutorials/beginner/saving_loading_models.html.

allaffa (approved):

PR #54, opened 2021-12-04
Adding profiling function

jychoi-hpc:
This is a PR for adding profiling routines based on PyTorch's Profiler.

### DIFF
--- a/hydragnn/train/train_validate_test.py
+++ b/hydragnn/train/train_validate_test.py
@@ -39,11 +39,14 @@ def train_validate_test(
     verbosity=0,
     plot_init_solution=True,
     plot_hist_solution=False,
+    profile=False,
 ):
     num_epoch = config["Training"]["num_epoch"]
     # setup profile variables
     global profile_config_name
     profile_config_name = model_with_config_name
+    global profile_enabled
+    profile_enabled = profile
     # total loss tracking for train/vali/test
     total_loss_train = []
     total_loss_val = []
@@ -199,9 +202,7 @@ def get_head_indices(model, data):
     return head_index
 
 def trace_handler(p):
-    rank = os.getenv("RANK")
     print ('Total number of profiled events: %d at epoch %d'%(len(p.events()), profile_current_epoch))
-    #p.export_chrome_trace("trace%s-%s-%d.json"%(rank, epoch, p.step_num))
     torch.profiler.tensorboard_trace_handler("./logs/" + profile_config_name)(p)
 
 def train(loader, model, opt, verbosity):
@@ -213,12 +214,11 @@ def train(loader, model, opt, verbosity):
 
     total_error = 0
     profiler = contextlib.nullcontext(MagicMock(name='step'))
-    if profile_current_epoch == 0:
+    if profile_enabled and (profile_current_epoch == 0):
         profiler = profile(
             activities=[ProfilerActivity.CPU, ProfilerActivity.CUDA],
             schedule=torch.profiler.schedule(
-                skip_first=10,
-                wait=5,
+                wait=1,
                 warmup=1,
                 active=3,
                 repeat=1),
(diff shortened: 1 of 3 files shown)


hydragnn/train/train_validate_test.py
```diff
@@ -35,8 +39,14 @@ def train_validate_test(
     verbosity=0,
     plot_init_solution=True,
     plot_hist_solution=False,
+    profile=False,
 ):
     num_epoch = config["Training"]["num_epoch"]
+    # setup profile variables
+    global profile_config_name
```
allaffa, line 46: @jychoi-hpc is there any easy-fix to avoid the use of global variables?

allaffa (changes requested):

hydragnn/train/train_validate_test.py
jychoi-hpc: Let me rewrite the code to remove global variables as much as possible.

jychoi-hpc:
I have rewritten the code. I added a new class(`Profiler` in `hydragnn/utils/profile.py`) to remove global variables.

hydragnn/train/train_validate_test.py
```diff
@@ -172,24 +186,38 @@ def get_head_indices(model, data):
...
-        loss, tasks_rmse = model.loss_rmse(pred, data.y, head_index)
-
-        loss.backward()
-        opt.step()
-        total_error += loss.item() * data.num_graphs
-        for itask in range(len(tasks_rmse)):
-            tasks_error[itask] += tasks_rmse[itask].item() * data.num_graphs
+    with profiler as prof:
+        for data in iterate_tqdm(loader, verbosity):
+            with record_function("load"):
```
streeve, line 205: We'll have to rebase on main - we no longer load here

streeve (commented):
This is a very nice added capability. I think I understand the basics, but is there any significant performance overhead (either with profiling turned on or off)?

It does make the training loop a little harder to read, but I think that's just the price we pay

hydragnn/train/train_validate_test.py
jychoi-hpc: Ok. Thank you for pointing this out. I am updating now.

hydragnn/train/train_validate_test.py
```diff
@@ -172,24 +186,34 @@ def get_head_indices(model, data):
...
-
-        pred = model(data)
-        loss, tasks_rmse = model.loss_rmse(pred, data.y, head_index)
-
-        loss.backward()
-        opt.step()
-        total_error += loss.item() * data.num_graphs
-        for itask in range(len(tasks_rmse)):
-            tasks_error[itask] += tasks_rmse[itask].item() * data.num_graphs
+    with profiler as prof:
```
allaffa, line 201: Is there a way to "hide" the action of the profiler so that someone who looks at the implementation of the training loop does not explicitly see the call tot he profile routines?

allaffa (changes requested):

hydragnn/train/train_validate_test.py
jychoi-hpc: Alternatively, we can put the profiler wrapper in one higher up:
```
    for epoch in range(0, num_epoch):
        profiler.set_current_epoch(epoch)
        with profiler as prof:
            train_rmse, train_taskserr = train(
                train_loader, model, optimizer, verbosity, profiler=prof
            )
        val_rmse, val_taskserr = validate(val_loader, model, verbosity)
```
Then, the train function will look like this:
```
def train(...):
    tasks_error = np.zeros(model.num_heads)
    model.train()
    total_error = 0
    for data in iterate_tqdm(loader, verbosity):
        with record_function("zero_grad"):
            opt.zero_grad()
        with record_function("get_head_indices"):
            head_index = get_head_indices(model, data)
        with record_function("forward"):
            pred = model(data)
            loss, tasks_rmse = model.loss_rmse(pred, data.y, head_index)
        with record_function("backward"):
            loss.backward()
        opt.step()
        prof.step()
```

jychoi-hpc: How do you think?

allaffa: @streeve @pzhanggit 
Which option do you prefer between the one already implemented and the new one outlined by Jong in this conversation?

allaffa: @jychoi-hpc 
I personally prefer the new proposed option. Let's see what Sam and Pei think :-)

streeve: I agree - I think this is ready to merge after the update

jychoi-hpc: I like the proposed option too. I have updated the code.

hydragnn/train/train_validate_test.py
```diff
@@ -172,21 +188,30 @@ def get_head_indices(model, data):
...
-        pred = model(data)
-        loss, tasks_rmse = model.loss_rmse(pred, data.y, head_index)
-
-        loss.backward()
+        with record_function("zero_grad"):
+            opt.zero_grad()
+        with record_function("get_head_indices"):
+            head_index = get_head_indices(model, data)
+        with record_function("forward"):
+            pred = model(data)
```
streeve, line 209: Should we separate the prediction from loss profiling?

streeve (approved):

allaffa (approved):

pzhanggit (approved):

PR #56, opened 2021-12-05
Utilities to compute formation enthalpy and Gibbs energy

allaffa:


streeve:
Still a draft or ready for review? This is a very helpful utility, but I think we need to split out the plotting and probably add as an option in the training script

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
```diff
@@ -0,0 +1,180 @@
...
+# directory.                                                                 #
+#                                                                            #
+# SPDX-License-Identifier: BSD-3-Clause                                      #
+##############################################################################
+
+import scipy.special
+import math
+import os
+import shutil
+import pandas
```
streeve, line 16: This will be a new dependency - probably okay to add a `requirements-utils.txt` with anything new here

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
```diff
@@ -0,0 +1,180 @@
...
+
+    plt.figure(2)
+    plt.scatter(composition_list, formation_gibbs_energy_list)
+    plt.xlabel("Fe concentration")
+    plt.ylabel("Formation Gibbs energy (Rydberg)")
+    plt.title("FePt")
+    plt.savefig("formation_gibbs_energy.png")
+
+
+def compute_formation_enthalpy(path_to_filename, elements_list, pure_elements_energy):
```
streeve, line 135: add a note that this is for binary only?

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
```diff
@@ -0,0 +1,180 @@
...
+        + (pure_elements_energy[element2] - pure_elements_energy[element1])
+        * (1 - composition_element1)
+    ) * num_atoms
+
+    formation_enthalpy = total_energy - linear_mixing_energy
+
+    return composition_element1, total_energy, linear_mixing_energy, formation_enthalpy
+
+
+if __name__ == "__main__":
```
streeve, line 177: Not sure we want this here - I think this call has to be embedded in the user's script

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
```diff
@@ -0,0 +1,180 @@
...
+
+    df = pandas.read_csv(path_to_filename, header=None, skiprows=1)
+    num_atoms = df.shape[0]
+
+    for atom_index in range(0, num_atoms):
+        row = df[0][atom_index].split()
+        # If just an int() cast is used in the following line, a runtime error occurs
+        # the following int(float()) cast is used to solve the runtime error
+        # https://stackoverflow.com/questions/1841565/valueerror-invalid-literal-for-int-with-base-10
+        atom_type_str = int(float(row[0]))
```
streeve, line 157: This fails for me locally unless I cast back to string

streeve (commented):
I'm not able to run with FePt - would this error clearly if I was missing the pure elements, for example?

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
```diff
@@ -0,0 +1,180 @@
...
+
+def from_energy_to_formation_gibbs(path_to_dir, elements_list, temperature_kelvin=0):
+
+    new_dataset_path = path_to_dir[:-1] + "_gibbs_energy/"
+
+    if os.path.exists(new_dataset_path):
+        shutil.rmtree(new_dataset_path)
+    os.makedirs(new_dataset_path)
+
+    Kb_joule_per_kelvin = 1.380649 * 1e-23
```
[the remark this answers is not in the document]
allaffa: @streeve 
The units cannot be changes in LSMS

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
allaffa: @streeve 
I made some calculation on paper: this policy works for any components. The problem is that the power of 10 used has to be larger than the total number of atoms in the lattice to avoid that two compositions are confused as the same.
Right now I am using 100 because the FePt dataset as 32 atoms. 
I already added a FIXME comment. 

Even if in exact arithmetic this policy always works, for obvious reasons you cannot extend its applicability in the code for a generic size. Numerical overflow will occur at some point. 
For example: 10,000 atoms in the lattice with 10 elements. Taking the power of 10,000 will very soon cause overflows. 

**I admit that I don't have a solution that would solve this problem for the general case.**

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
```diff
@@ -0,0 +1,180 @@
...
+def get_key(dictionary, val):
+
+    for key, value in dictionary.items():
+        if val == value:
+            return key
+
+    return None
+
+
+def from_energy_to_formation_gibbs(path_to_dir, elements_list, temperature_kelvin=0):
```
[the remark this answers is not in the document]
allaffa: `convert_total_energy_to_enthalpy`

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
allaffa: Do you mean that also the pytests fail for you locally?

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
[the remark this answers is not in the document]
allaffa: For a temperature far from zero, the function computed the Gibbs energy

G = H - TS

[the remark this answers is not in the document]
streeve: Right, how about `convert_raw_energy_to_gibbs`. I want it to be clear that this is operating on raw data files

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
streeve: Unit test passed (which confuses me more)

streeve: Ok, I did get it to run with `utils.lsms.convert_total_energy_to_formation_gibbs("/Users/5t2/Downloads/FePt/", ['78', '26'])`. I was not previously string quoting the element numbers

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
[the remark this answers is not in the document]
allaffa: Can you take care fo this small changes and then merge the PR? Thanks.

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
allaffa: user@localadmins-Air HydraGNN % grep -n -r 'import os' ./
**.//hydragnn/preprocess/raw_dataset_loader.py:12:import os
.//hydragnn/preprocess/load_data.py:12:import os
.//hydragnn/utils/model.py:12:import os
.//hydragnn/utils/distributed.py:12:import os
.//hydragnn/utils/config_utils.py:12:import os
.//hydragnn/models/create.py:12:import os
.//hydragnn/train/train_validate_test.py:25:import os
.//setup.py:1:import os**

`os` should be in `requirements.txt`

allaffa: Thanks. Done.

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
allaffa: Thanks. Done.

allaffa:
@streeve @pzhanggit 
Let me know if there is anything else to address in this PR

utils/lsms/convert_total_energy_to_formation_gibbs.py
```diff
@@ -0,0 +1,186 @@
...
+import scipy.special
+import math
+import os
+import shutil
+import pandas
+import numpy as np
+import matplotlib.pyplot as plt
+
+# function to return key for any value
+def get_key(dictionary, val):
```
jychoi-hpc, line 21: If dictionary is large, this function can be slow. You may want to change to something like:
```
if key in dictionary:
    return dictionary[key]
else:
    return None
```

jychoi-hpc: Sorry. My misunderstanding. It is trying to find value.

streeve:
I have some local updates for this, but need #93 for the test changes - @allaffa are you okay if I rebase?

tests/deterministic_graph_data.py
```diff
@@ -52,83 +103,68 @@ def deterministic_graph_data(
...
+                positions[count_pos][1] = y
+                positions[count_pos][2] = z
+                positions[count_pos + 1][0] = x + 0.5
+                positions[count_pos + 1][1] = y + 0.5
+                positions[count_pos + 1][2] = z + 0.5
+                count_pos = count_pos + 2
+
+    node_ids = torch.tensor([int(i) for i in range(0, number_nodes)]).reshape(
+        (number_nodes, 1)
     )
```
pzhanggit, line 122:
```suggestion
    node_ids = torch.tensor(range(number_nodes), dtype=torch.int64).reshape(
            (number_nodes, 1)
    )
```

pzhanggit (approved):
Looks good to me.

tests/deterministic_graph_data.py
streeve: Fixed

pzhanggit: Thanks. One minor thing about the for loops line 109-118, they can be replaced/optimized with tensor operations. But I am also okay with update it later in a different PR.

tests/deterministic_graph_data.py
```diff
@@ -52,83 +103,68 @@ def deterministic_graph_data(
...
-    number_unit_cell_x = torch.randint(
-        number_unit_cell_x_range[0],
-        number_unit_cell_x_range[1],
-        (number_configurations,),
-    )
-    number_unit_cell_y = torch.randint(
-        number_unit_cell_y_range[0],
-        number_unit_cell_y_range[1],
-        (number_configurations,),
+    count_pos = 0
```
pzhanggit, line 106:
```suggestion
```

tests/deterministic_graph_data.py
```diff
@@ -52,83 +103,68 @@ def deterministic_graph_data(
...
+    for x in range(uc_x):
+        for y in range(uc_y):
+            for z in range(uc_z):
+                positions[count_pos][0] = x
+                positions[count_pos][1] = y
+                positions[count_pos][2] = z
+                positions[count_pos + 1][0] = x + 0.5
+                positions[count_pos + 1][1] = y + 0.5
+                positions[count_pos + 1][2] = z + 0.5
+                count_pos = count_pos + 2
```
pzhanggit, line 118:
```suggestion
    x = torch.range(0, uc_x - 1)
    y = torch.range(0, uc_y - 1)
    z = torch.range(0, uc_z - 1)
    x3d, y3d, z3d = torch.meshgrid(x, y, z)
    positions[::2, 0] = x3d.reshape(-1, 1)
    positions[::2, 1] = y3d.reshape(-1, 1)
    positions[::2, 2] = z3d.reshape(-1, 1)
    positions[1::2, :] = positions[::2, :] + 0.5
```

streeve: This looks great, but can we address separately like you mention? I don't really follow and don't want to hold this up any longer

pzhanggit: Sure.

utils/LSMS_utils/convert_total_energy_in_gibbs_energy.py
streeve: Max fixed this in later updates

PR #59, opened 2021-12-08
Clean up error naming and list issues 

pzhanggit:
- [x] Fix incorrect "MAE", "MSE" naming to "RMSE" 
- [x] Remove error list may not be needed, e.g., error of sum, anymore
- [x] Remove list append

### DIFF
--- a/hydragnn/train/train_validate_test.py
+++ b/hydragnn/train/train_validate_test.py
@@ -41,10 +41,6 @@ def train_validate_test(
     total_loss_train = []
     total_loss_val = []
     total_loss_test = []
-    # loss tracking of summation across all nodes for node feature predictions
-    task_loss_train_sum = []
-    task_loss_test_sum = []
-    task_loss_val_sum = []
     # loss tracking for each head/task
     task_loss_train = []
     task_loss_test = []
@@ -70,9 +66,9 @@ def train_validate_test(
     visualizer.num_nodes_plot()
 
     if plot_init_solution:  # visualizing of initial conditions
-        test_res = test(test_loader, model, verbosity)
-        true_values = test_res[3]
-        predicted_values = test_res[4]
+        test_results = test(test_loader, model, verbosity)
+        true_values = test_results[2]
+        predicted_values = test_results[3]
         visualizer.create_scatter_plots(
             true_values,
             predicted_values,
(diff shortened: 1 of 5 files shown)


hydragnn/train/train_validate_test.py
```diff
@@ -70,9 +70,9 @@ def train_validate_test(
     visualizer.num_nodes_plot()
 
     if plot_init_solution:  # visualizing of initial conditions
-        test_rmse = test(test_loader, model, verbosity)
-        true_values = test_rmse[3]
-        predicted_values = test_rmse[4]
+        test_res = test(test_loader, model, verbosity)
```
streeve, line 73: Why 'res'?

streeve:
Replacing list appending seems like a big job - separate PR?

pzhanggit:
> Replacing list appending seems like a big job - separate PR?

I am just going to replace the appending in error lists in this PR.

hydragnn/train/train_validate_test.py
streeve: Got it. `test_results`, please :)

pzhanggit: Done.

streeve:
So to be clear, we just don't care about the node sum error anymore?

pzhanggit:
> So to be clear, we just don't care about the node sum error anymore?

Yes, especially given the good results @allaffa got. At one point, we can only get good predictions for sum of node values and had the discussion about loss of node sum. But I do not think we need it any more.

hydragnn/train/train_validate_test.py
```diff
@@ -70,9 +66,9 @@ def train_validate_test(
     visualizer.num_nodes_plot()
 
     if plot_init_solution:  # visualizing of initial conditions
-        test_rmse = test(test_loader, model, verbosity)
-        true_values = test_rmse[3]
-        predicted_values = test_rmse[4]
+        test_results = test(test_loader, model, verbosity)
```
streeve, line 69: Maybe we just return all the values separately here too? I think will be easier to read

pzhanggit: Done.

streeve (approved):
Looks good to me

streeve:
@allaffa okay to merge?

allaffa (approved):

PR #63, opened 2021-12-14
added dashed red diagonal on scatterplot

allaffa:
Also makes scatter plot empty markers

### DIFF
--- a/hydragnn/postprocess/visualizer.py
+++ b/hydragnn/postprocess/visualizer.py
@@ -113,7 +113,7 @@ class Visualizer:
         y_label=None,
         xylim_equal=False,
     ):
-        ax.scatter(x, y, s=s, edgecolor='b', marker=marker, facecolor="none")
+        ax.scatter(x, y, s=s, edgecolor="b", marker=marker, facecolor="none")
 
         ax.set_title(title)
         ax.set_xlabel(x_label)
--- a/hydragnn/postprocess/visualizer.py
+++ b/hydragnn/postprocess/visualizer.py
@@ -113,7 +113,8 @@ class Visualizer:
         y_label=None,
         xylim_equal=False,
     ):
-        ax.scatter(x, y, s, c, marker)
+        ax.scatter(x, y, s=s, edgecolor='b', marker=marker, facecolor="none")
+
         ax.set_title(title)
         ax.set_xlabel(x_label)
         ax.set_ylabel(y_label)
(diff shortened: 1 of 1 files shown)


hydragnn/postprocess/visualizer.py
```diff
@@ -550,6 +553,21 @@ def create_scatter_plot_nodes_vec(
                 fig.savefig(f"./logs/{self.model_with_config_name}/" + varname + ".png")
             plt.close()
 
+    def add_identity(self, axes, *line_args, **line_kwargs):
+        (identity,) = axes.plot([], [], *line_args, **line_kwargs)
+
+        def callback(axes):
```
streeve, line 559: Not sure I understand why the callback is needed

hydragnn/postprocess/visualizer.py
```diff
@@ -550,6 +553,21 @@ def create_scatter_plot_nodes_vec(
                 fig.savefig(f"./logs/{self.model_with_config_name}/" + varname + ".png")
             plt.close()
 
+    def add_identity(self, axes, *line_args, **line_kwargs):
```
streeve, line 556: Is it bad to just plot a line `y=x`? I'm not familiar with the approach here

allaffa:
@streeve 
For more details:
https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.add_callback.html

hydragnn/postprocess/visualizer.py
streeve: Makes sense. I tried it out - very fancy. I'll open an issue to make sure you can actually interact with plots rather than only save them

streeve (approved):

hydragnn/postprocess/visualizer.py
```diff
@@ -124,6 +125,8 @@ def __scatter_impl(
             ax.set_xlim(minimum, maximum)
             ax.set_ylim(minimum, maximum)
 
+        self.add_identity(ax, color="r", ls="--")
```
pzhanggit, line 128: Move this inside ` if xylim_equal:`?

pzhanggit (commented):
Looks good to me. Only one comment about moving the identity plot inside if

hydragnn/postprocess/visualizer.py
streeve: No, I think we want this regardless of forcing x_lim = y_lim

PR #65, opened 2021-12-15
stratified splitting

allaffa:


hydragnn/preprocess/load_data.py
```diff
@@ -128,6 +147,103 @@ def combine_and_split_datasets(
 
     return train_loader, val_loader, test_loader
 
+def stratified_splitting(dataset, perc_train):
```
allaffa, line 150: @streeve I played making counts on a piece of paper, and I think tat this splitting works for a generic multi-component system. :-)

allaffa:
@streeve  @pzhanggit 

DO NOT review. I converted it to draft. I need to add a unit test first.

hydragnn/preprocess/load_data.py
```diff
@@ -129,6 +151,94 @@ def combine_and_split_datasets(
     return train_loader, val_loader, test_loader
 
 
+def stratified_splitting(dataset, perc_train):
```
allaffa, line 154:
```suggestion
def stratified_splitting(dataset, perc_train):
#FIXME: this routine works only if the total number of atoms is below 100
```

allaffa: @streeve 
If the total number of atoms is below 1000, then this idea works
In general, you need to replace "100" in this routing with the smallest power of 10 that exceeds the Toal number of atoms.

hydragnn/preprocess/load_data.py
```diff
@@ -25,19 +25,36 @@
...
+# function to return key for any value
+def get_keys(dictionary, val):
+    keys = []
+    for key, value in dictionary.items():
+        if val == value:
+            keys.append(key)
+    return keys
+
+
+def dataset_loading_and_splitting(config: {}):
```
pzhanggit, line 41: Not rebased properly? These have been merged into main and should not show up as new changes.

allaffa: I will delete this branch and open a new one from scratch.

PR #66, opened 2021-12-15
save train/val/test pk files when raw total provided

pzhanggit:
- [x] Save *train.pkl, *validate.pkl and *test.pkl files used in training when total dataset is provided for consistency in continuing training, comparison etc.
- [x] remove "raw" since not needed

### DIFF
--- a/hydragnn/preprocess/load_data.py
+++ b/hydragnn/preprocess/load_data.py
@@ -25,19 +25,26 @@ from hydragnn.preprocess.serialized_dataset_loader import SerializedDataLoader
 from hydragnn.preprocess.raw_dataset_loader import RawDataLoader
 from hydragnn.utils.distributed import get_comm_size_and_rank
 from hydragnn.utils.time_utils import Timer
+import pickle
 
 
-def dataset_loading_and_splitting(
-    config: {},
-    chosen_dataset_option,
-):
+def dataset_loading_and_splitting(config: {}):
 
-    dataset_chosen, dataset_names = load_data(chosen_dataset_option, config)
-    return split_dataset(
-        dataset_list=dataset_chosen,
-        dataset_names=dataset_names,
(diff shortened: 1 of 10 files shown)


pzhanggit:
@allaffa  See this PR for separate pkl files

hydragnn/preprocess/load_data.py
```diff
@@ -171,8 +175,27 @@ def transform_raw_data_to_serialized(config):
     _, rank = get_comm_size_and_rank()
 
     if rank == 0:
-        loader = RawDataLoader(config)
+        loader = RawDataLoader(config["Dataset"])
         loader.load_raw_data()
+        if len(loader.serial_data_name_list) == 1 and "total" in loader.path_dictionary:
```
streeve, line 180: Let's move everything in this `if` into a separate function

hydragnn/preprocess/load_data.py
```diff
@@ -76,22 +82,24 @@ def create_dataloaders(trainset, valset, testset, batch_size):
     return train_loader, val_loader, test_loader
 
 
-def split_dataset(
-    dataset_list: [],
-    dataset_names: [],
-    batch_size: int,
-    perc_train: float,
-):
+def split_dataset(dataset: [], perc_train: float):
```
streeve, line 85: I like this cleanup, but you should add an issue for cleaning up the function names

allaffa: @streeve 
Do you mean open an issue on GutHub and link it to this PR?

streeve: That was my intent, yes. I should have really said clean up naming and intent of each function - it's always hard for me to remember what each function here actually does

pzhanggit:
I updated the script after our discussion. Let me know what you think @streeve

allaffa (approved):

PR #73, opened 2021-12-16
Move data to device from loaders

streeve:
DO NOT MERGE THIS HAS NOT BEEN FULLY TESTED
Make sure batch variables in each data sample are moved to the device


streeve:
This needs to be fixed - I only move to the device when using DDP right now

PR #76, opened 2021-12-17
Add `precommit`

### LINKED ISSUE #75: Add pre-commit to make contributing easy.
I noticed that in the [`CONTRIBUTING.md`](https://github.com/ORNL/HydraGNN/blob/main/CONTRIBUTING.md) mentions that the contributor must run `black` and `pytest` before contributing. Having [`pre-commit`](https://pre-commit.com/) would make that process much easier and enforce that the developers follow coding conventions.

I'd be more than happy to take this up.

SauravMaheshkar:
Fixes #75.

Based on the feedback provided by @streeve I've added `black` among some other miscellaneous checks to the `.pre-commit-config.yaml`. I've also added `pre-commit` to the `requirements-dev.txt` file in order to aid new contributors.

I have not added `pytest` as of right now based on the feedback provided but more than happy to take that up again. 

I'd also recommend adding instructions on how to install pre-commit hooks to the [**`CONTRIBUTING.md`**](https://github.com/ORNL/HydraGNN/blob/main/CONTRIBUTING.md) file.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,8 +5,8 @@ repos:
       - id: check-merge-conflict
       - id: trailing-whitespace
       - id: end-of-file-fixer
-  
+
   - repo: https://github.com/psf/black
-    rev: 19.3b0
+    rev: 21.5b1
     hooks:
-      - id: black 
+      - id: black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,12 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.4.0
+    hooks:
+      - id: check-merge-conflict
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+  
+  - repo: https://github.com/psf/black
+    rev: 19.3b0
+    hooks:
+      - id: black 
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+repos:
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v3.4.0
+    hooks:
+      - id: check-merge-conflict
+      - id: trailing-whitespace
+      - id: end-of-file-fixer
+
+  - repo: https://github.com/psf/black
+    rev: 21.5b1
```
streeve, line 10: I changed this just to match the current version enforced in the CI

streeve (approved):
This is great! Thanks for contributing. If there is a way to add `pytest`, but turn it off by default, I think it would also be very useful. But that does not necessarily need to be done in the same PR

.pre-commit-config.yaml
SauravMaheshkar: There's the option of turning off git hooks with `git commit --no-verify -m '<insert-commit-message>'`. But that's a hack IMO. I didn't `pytest` because I noticed that the repository already has CI that runs on every PR which also runs `pytest`.

SauravMaheshkar:
PS: Apparently there are some files which are corrected by `black`. Maybe I can do a follow up PR ? This PR would just contain the changes after running black on all files.

allaffa (approved):

PR #77, opened 2021-12-17
Simplify output name

streeve:
Small fixup and simplification. Output name only needs config, not model

### DIFF
--- a/hydragnn/run_prediction.py
+++ b/hydragnn/run_prediction.py
@@ -21,8 +21,8 @@ from hydragnn.utils.model import load_existing_model
 from hydragnn.utils.time_utils import print_timers
 from hydragnn.utils.config_utils import (
     update_config_NN_outputs,
-    get_model_output_name_config,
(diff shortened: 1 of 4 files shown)


hydragnn/run_prediction.py
```diff
@@ -81,8 +81,8 @@ def _(config: dict):
         verbosity=config["Verbosity"]["level"],
     )
 
-    model_with_config_name = get_model_output_name_config(model, config)
-    load_existing_model(model, model_with_config_name)
+    output_name = get_output_name_config(config)
```
pzhanggit, line 84: Maybe a different name, like "model_output_name" or "model_config_name"? "output_name" may be a little confusing with `config["Variables_of_interest"]["output_names"]`?

streeve: You're right. I was trying to avoid `model_name` because this is the path for all output though.

pzhanggit (approved):

PR #80, opened 2021-12-18
optimize get_head_indices with tensor operations

pzhanggit:
Tested on Summit with profiling in https://github.com/ORNL/HydraGNN/pull/54, about 15x speedup is observed

### DIFF
--- a/hydragnn/preprocess/serialized_dataset_loader.py
+++ b/hydragnn/preprocess/serialized_dataset_loader.py
@@ -128,7 +128,9 @@ class SerializedDataLoader:
             A Data object representing a structure that has atoms.
         """
         output_feature = []
-        data.y_loc = torch.zeros(1, len(type) + 1, dtype=torch.int64)
+        data.y_loc = torch.zeros(
+            1, len(type) + 1, dtype=torch.int64, device=data.y.device
+        )
         for item in range(len(type)):
             if type[item] == "graph":
                 feat_ = torch.reshape(data.y[index[item]], (1, 1))


hydragnn/train/train_validate_test.py
```diff
@@ -160,14 +160,25 @@ def get_head_indices(model, data):
     y_loc = data.y_loc
     # head size for each sample
     total_size = y_loc[:, -1]
-    head_index = []
+    # feature index for all heads
+    head_index = [None] * model.num_heads
+    # intermediate work list
+    _head_ind = [None] * batch_size
```
streeve, line 166: Can we rename this variable to something more clear? Even just something saying it's temporary

hydragnn/train/train_validate_test.py
```diff
@@ -160,14 +160,25 @@ def get_head_indices(model, data):
...
+    # track the start loc of each sample
+    sample_start = torch.cumsum(total_size, dim=0) - total_size
+    sample_start = sample_start.view(-1, 1)
+    # shape (batch_size, model.num_heads), start and end of each head for each sample
+    start_track = sample_start + y_loc[:, :-1]
```
streeve, line 171: I don't understand "track". `start_index`?

hydragnn/train/train_validate_test.py
```diff
@@ -160,14 +160,25 @@ def get_head_indices(model, data):
...
+    end_track = sample_start + y_loc[:, 1:]
+
+    # a large index tensor pool for all element in data.y
+    index_pool = torch.arange(0, end_track[-1, -1])
```
streeve, line 175: `index_range`?

streeve (changes requested):
It will be great to have this merged! Just requesting name changes, as usual

jychoi-hpc:
I am just wondering, if we make `head_index` as a torch array on GPU, will it be better? 

It can get some speed up when it used with other tensor data on GPUs, e.g.,:
```
loss, tasks_rmse = model.loss_rmse(pred, data.y, head_index)
```
`pred`, `data.y` are all on GPU, right?

pzhanggit:
> I am just wondering, if we make `head_index` as a torch array on GPU, will it be better?
> 
> It can get some speed up when it used with other tensor data on GPUs, e.g.,:
> 
> ```
> loss, tasks_rmse = model.loss_rmse(pred, data.y, head_index)
> ```
> 
> `pred`, `data.y` are all on GPU, right?

Good point. Currently it is a list of tenors. `head_index=[tensor_index_head0, tensor_index_head1,...]`.  The thing is that the length of `head_index` for each head is different, for example 32 for `tensor_index_head0` and 32*num_nodes for `tensor_index_head1` when batch_size=32. Not sure how to define a tensor for this, maybe padding? What do you think?
I will make the `tensor_index_head*` on GPU.

jychoi-hpc:
So, the each element in `head_index` in a tensor (on GPU, if GPU is available). Then, it should be ok with `list`.

pzhanggit:
> So, the each element in `head_index` in a tensor (on GPU, if GPU is available). Then, it should be ok with `list`.

I will make it so explicitly in a new commit.

hydragnn/train/train_validate_test.py
```diff
@@ -160,14 +160,25 @@ def get_head_indices(model, data):
     y_loc = data.y_loc
     # head size for each sample
     total_size = y_loc[:, -1]
-    head_index = []
+    # feature index for all heads
+    head_index = [None] * model.num_heads
+    # intermediate work list
+    _head_ind = [None] * batch_size
+    # track the start loc of each sample
+    sample_start = torch.cumsum(total_size, dim=0) - total_size
```
streeve, line 168: Along with what @jychoi-hpc was saying, I think you can specify the device (same for `arange`):
`sample_start = torch.cumsum(total_size, dim=0, out=torch.Tensor(device=device))`

pzhanggit: Yeah, did so. Waiting for nodes on Summit for testing.

pzhanggit: @streeve  @jychoi-hpc  The tensors should be in the same device with data.x and data.y now.

jychoi-hpc (approved):
It looks ok to me.

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -128,7 +128,9 @@ def __update_predicted_values(self, type: list, index: list, data: Data):
             A Data object representing a structure that has atoms.
         """
         output_feature = []
-        data.y_loc = torch.zeros(1, len(type) + 1, dtype=torch.int64)
+        data.y_loc = torch.zeros(
+            1, len(type) + 1, dtype=torch.int64, device=data.y.device
```
streeve, line 132: This is a very nice way to reduce calls to `get_device` and simplify

hydragnn/train/train_validate_test.py
streeve: Ok, since you moved `y_loc` to the device, this will default there as well?

streeve (approved):
Looks great!

hydragnn/train/train_validate_test.py
pzhanggit: Yes.

PR #81, opened 2021-12-19
Stratified splitting

allaffa:


tests/inputs/ci.json
```diff
@@ -5,6 +5,7 @@
     "Dataset": {
         "name": "unit_test_singlehead",
         "format": "unit_test",
+        "stratified_splitting": true,
```
pzhanggit, line 8: `stratified_splitting` will not be executed in this test since train/test/val sets are provided separately.

allaffa: `stratified_splitting` should be called before the creation of the .pkl files. 

I agree that the unit tests will use the .pkl files once they are created, thus avoiding re-generating the raw data from scratch. But the goal it so stratify the splitting the first time, so that the creation go the p.kl files ensure an equal representation of all the compositions in the training/validate/test portion of the dataset. 

What am I missing?

pzhanggit: Nothing is missing. Just a comment `StratifiedShuffleSplit` is not used even though set to be `true` here in the singletask test since the datasets are provided separately in `unittest_train_model`.

allaffa: I see...
Do you think it would be good if I added the use of `StratifiedShuffleSplit` inside `unittest_train_model`

allaffa: @pzhanggit 
I understand what's happening. What you say is correct. 
I removed the `stratified_splitting` from the `ci.json` configure file

hydragnn/preprocess/load_data.py
```diff
@@ -83,15 +93,108 @@ def create_dataloaders(trainset, valset, testset, batch_size):
...
+        Subsample of the original dataset constructed using stratified sampling.
+    """
+    dataset_categories = []
+
+    for data in dataset:
+        frequencies = torch.bincount(data.x[:, 0].int())
+        frequencies = sorted(frequencies[frequencies > 0].tolist())
+        category = 0
+        for index, frequency in enumerate(frequencies):
+            category += frequency * (100 ** index)
```
pzhanggit, line 139: Is this representation unique? For example, a structure with 4 Fe atoms and 6 Pt atoms would have the same `category` value (=4+6*100?) with a different structure with 6 Fe and 4 Pt?

hydragnn/preprocess/load_data.py
```diff
@@ -83,15 +93,108 @@ def create_dataloaders(trainset, valset, testset, batch_size):
...
+        Percentage of the dataset.
+    Returns
+    ----------
+    [Data]
+        Subsample of the original dataset constructed using stratified sampling.
+    """
+    dataset_categories = []
+
+    for data in dataset:
+        frequencies = torch.bincount(data.x[:, 0].int())
```
pzhanggit, line 135: Is `stratified_splitting` function specifically designed for cases that `data.x[:, 0]` are integers or can be uniquely represented by integers? For example, this line would fail to tell the different structures in unit tests, where`data.x[:,0]` is not greater than 1 for all nodes.

pzhanggit: Btw, even for `lsms` data, `data.x[:,0]` is normalized between 0 and 1 at this point. So, the current counting works for binary alloy, i.e., with 0 and 1 `data.x[:,0]`, for but would fail for more than two atom types?

pzhanggit (changes requested):
Not sure current `category` value is a unique representation of data structures (see the detailed comments.)

hydragnn/preprocess/load_data.py
```diff
@@ -83,15 +93,108 @@ def create_dataloaders(trainset, valset, testset, batch_size):
...
+
 
-    perc_val = (1 - perc_train) / 2
-    data_size = len(dataset)
-    trainset = dataset[: int(data_size * perc_train)]
-    valset = dataset[
-        int(data_size * perc_train) : int(data_size * (perc_train + perc_val))
-    ]
-    testset = dataset[int(data_size * (perc_train + perc_val)) :]
+def stratified_splitting(dataset, perc_train):
```
streeve, line 116: Isn't this really `compositional_stratified_splitting`? It probably makes sense to have another `stratified_splitting` function that just wraps sklearn

hydragnn/preprocess/load_data.py
allaffa: > Is this representation unique? For example, a structure with 4 Fe atoms and 6 Pt atoms would have the same `category` value (=4+6*100?) with a different structure with 6 Fe and 4 Pt?

You are right. The representation is not unique. I need to fix this.

hydragnn/preprocess/load_data.py
allaffa: Yes, `compositional_stratified_splitting` is a more appropriate term. 
As to to having another` stratified_splitting` function that just wraps sklearn, do you have in mind anything specific?

hydragnn/preprocess/load_data.py
allaffa: @pzhanggit 
After the discussion we had to today where I explained to you the rationale behind the new categorization, let me know if the new version of the code addresses your comment.

hydragnn/preprocess/load_data.py
allaffa: @pzhanggit 
I use a dictionary now to map different elements rescaled proton numbers into integers.

hydragnn/preprocess/load_data.py
allaffa: > Isn't this really `compositional_stratified_splitting`? It probably makes sense to have another `stratified_splitting` function that just wraps sklearn

@streeve 
I changed the name of the function to `compositional_stratified_splitting` as you suggested

tests/inputs/ci.json
allaffa: @streeve 
Taken out :)

hydragnn/preprocess/load_data.py
```diff
@@ -83,15 +94,142 @@ def create_dataloaders(trainset, valset, testset, batch_size):
...
+        elements_set = torch.cat((elements_set, elements), 0)
+
+    elements_set = torch.unique(elements_set)
+    dictionary = {}
+    power_ten = math.ceil(math.log10(max_graph_size))
+
+    for index, element in enumerate(list(elements_set)):
+        dictionary[element.item()] = index
+
+    for data in dataset:
```
pzhanggit, line 153: Wrap these lines into functions since they are used repeatedly below to generate val and test sets also

allaffa: Done. :) Thanks.

hydragnn/preprocess/load_data.py
```diff
@@ -27,9 +29,106 @@
...
+# function to return key for any value
+def get_keys(dictionary, val):
+    keys = []
+    for key, value in dictionary.items():
+        if val == value:
+            keys.append(key)
+    return keys
+
+
+def get_max_graph_size(dataset: torch_geometric.data.dataset):
```
pzhanggit, line 43: What is the type `torch_geometric.data.dataset`?

hydragnn/preprocess/load_data.py
```diff
@@ -27,9 +29,106 @@
...
+
+    return dataset, dataset_categories
 
-def dataset_loading_and_splitting(config: {}):
 
+def generate_partition(
+    sss: sklearn.model_selection.StratifiedShuffleSplit, dataset, dataset_categories
+):
+    parition1_indices = []
+    partition2_indices = []
```
pzhanggit, line 114:
```suggestion
```

hydragnn/preprocess/load_data.py
```diff
@@ -27,9 +29,106 @@
...
+        max_graph_size = max(max_graph_size, data.num_nodes)
+    return max_graph_size
+
+
+def get_elements_set(dataset: torch_geometric.data.dataset):
+    ## Identify all the elements present in at least one configuration
+    elements_set = torch.FloatTensor()
+
+    for data in dataset:
+        elements = torch.unique(data.x[:, 0])
```
pzhanggit, line 55: Currently `dataset_categories` is hard coded to be calculated by` data.x[:, 0]`. Could we make it optional? We might want to use a different variable for stratified splitting.

pzhanggit (approved):
Thanks for the changes. It is looks great to me now. Only one question about the variable used to stratified stratification, should it be optional?

hydragnn/preprocess/load_data.py
allaffa: A few notes on this point: 
 
1) The quantity can be either an input or an output
2) the user has to specify where the quantity can be found the data (i.e. the index of the column). Different input formats impact this decision. To circumvent this, the choice of the labeling quantity has to happen on torch.geometric.data, meaning after the quantities are imported
3) Related to point 2, the user DOES not have direct knowledge of what the labeling quantity si going to end up in the torch.geometric.structure
4) The quantity has to be a scalar. I don't see how the splitting could be done using vectors as labels
5) The quantity used to split is not necessarily something we need to use also as a feature for predictions

PR #84, opened 2021-12-25
citation file added

allaffa:


jychoi-hpc (approved):
It looks good to me.

streeve (approved):
As I understand, this `cff` format only allows one citation. So in the future we can add articles to the readme and/or replace here as we want

PR #86, opened 2022-01-05
Remove num_nodes_list

streeve:
This simplifies creating the histogram of nodes per graph, by using built-in `num_nodes` instead. Blocks #35


pzhanggit (approved):
Thanks.

PR #87, opened 2022-01-11
Ising model

allaffa:


allaffa:
This is the performance I get by training a HydraGNN model with 

**number_atoms_per_dimension = 5
configurational_histogram_cutoff = 100**


total_energy
magnetic_moment

examples/ising_model/ising_model.json
```diff
@@ -0,0 +1,57 @@
+{
+    "Verbosity": {
+        "level": 2
+    },
+    "Dataset": {
+        "name": "ising_model_27atoms",
+        "path": {"total": "./dataset/ising_model_data"},
+        "format": "unit_test",
+        "node_features": {
+            "name": ["atom_input_feature","magnetic_moment"],
```
streeve, line 10: "atom_type" and "spin"?

examples/ising_model/requirements.txt
```diff
@@ -0,0 +1 @@
+sympy
```
streeve, line 1: Not really sure of convention for this type of case. Can either leave it here or move to top level, `requirements-examples`

examples/ising_model/create_configurations.py
```diff
@@ -0,0 +1,109 @@
+import os
+import shutil
+import numpy as np
+from sympy.utilities.iterables import multiset_permutations
+import scipy.special
+
+
+def write_to_file(total_energy, atomic_features, count_config, dir):
```
streeve, line 8: I wonder if we want this file in the top-level `utils/`. Although I like that these files are all together

examples/ising_model/create_configurations.py
```diff
@@ -0,0 +1,109 @@
...
+                    + config[x, (y + 1) % L, z]
+                    + config[(x - 1) % L, y, z]
+                    + config[x, (y - 1) % L, z]
+                    + config[x, y, z]
+                    + config[x, y, (z + 1) % L]
+                    + config[x, y, (z - 1) % L]
+                )
+                total_energy += -nb * S
+
+                atomic_features[count_pos, 0] = config[x, y, z]
```
streeve, line 53: Do we still want this to be "type", position, and spin? Looks like spin and spin right now

examples/ising_model/create_configurations.py
```diff
@@ -0,0 +1,109 @@
...
+                total_energy, atomic_features = E_dimensionless(config, L)
+
+                write_to_file(total_energy, atomic_features, count_config, dir)
+
+                count_config = count_config + 1
+
+
+if __name__ == "__main__":
+
+    dir = "../../dataset/ising_model_data"
```
streeve, line 101: We can make sure it goes where we intend: `dir = os.path.join(os.path.dirname(__file__), "../../dataset/ising_model")`

examples/ising_model/create_configurations.py
```diff
@@ -0,0 +1,109 @@
...
+    total_energy = total_energy / 6
+
+    return total_energy, atomic_features
+
+
+def create_dataset(L, histogram_cutoff, dir):
+
+    count_config = 0
+
+    for num_downs in range(0, L ** 3):
```
streeve, line 68: I added `tqdm` here when I tested :)

examples/ising_model/ising_model.json
```diff
@@ -0,0 +1,57 @@
+{
+    "Verbosity": {
+        "level": 2
+    },
+    "Dataset": {
+        "name": "ising_model_27atoms",
+        "path": {"total": "./dataset/ising_model_data"},
```
streeve, line 7: dataset/ising_model is probably fine

streeve (commented):
This looks great! I was also able to start testing locally

examples/ising_model/ising_model.json
allaffa: yes, good point

examples/ising_model/requirements.txt
allaffa: On another repo, Miro was suggesting to have the requirements for specific example at the same directory level as the example itself. So I would keep it here. 
Do you still want me to rename the file or leave the name as is?

examples/ising_model/create_configurations.py
allaffa: As for now, this is used only here. Right? 
In case we are using it somewhere else later, I would move it. But for the time being, if this is specific to this example, I would keep it here.

examples/ising_model/create_configurations.py
allaffa: Can we explain what this line does?

examples/ising_model/requirements.txt
streeve: I'm okay with it as is

examples/ising_model/create_configurations.py
streeve: Adds the relative path to the `dataset` directory to where this file exists instead of where the user runs. If you run the original line with `python examples/ising_model/create_configurations.py` it will put the data outside hydra

examples/ising_model/create_configurations.py
streeve: Done

examples/ising_model/ising_model.json
streeve: Also done

streeve (approved):

examples/ising_model/create_configurations.py
```diff
@@ -0,0 +1,110 @@
...
+    total_energy = 0
+
+    count_pos = 0
+    number_nodes = L ** 3
+    positions = np.zeros((number_nodes, 3))
+    atomic_features = np.zeros((number_nodes, 5))
+    for z in range(L):
+        for y in range(L):
+            for x in range(L):
+                positions[count_pos, 0] = x - 1
```
pzhanggit, line 38: Why `-1`?

pzhanggit:
> This is the performance I get by training a HydraGNN model with
> 
> **number_atoms_per_dimension = 5 configurational_histogram_cutoff = 100**
> 
> total_energy magnetic_moment

@allaffa @streeve  Does it make sense to view spin prediction as a classification problem in the Ising model?

examples/ising_model/create_configurations.py
allaffa: I removed it. Indeed, it was not needed.

allaffa:
> 

@pzhanggit it does make sense, but I would like to give the idea that the Ising model naturally flows into the FePt case without changing of predictive tasks.

PR #89, opened 2022-01-14
Make Ising example configurable

streeve:
Makes the spin value randomization and non-linear spin function optional. With this update running the function with defaults results in the original Ising model

### DIFF
--- a/examples/ising_model/create_configurations.py
+++ b/examples/ising_model/create_configurations.py
@@ -26,10 +26,13 @@ def write_to_file(total_energy, atomic_features, count_config, dir):
 
 
 # 3D Ising model
-def E_dimensionless(config, L, spin_function):
+def E_dimensionless(config, L, spin_function, scale_spin):
     total_energy = 0
-
     spin = np.zeros_like(config)
+
+    if scale_spin:
+        random_scaling = np.random.random((L, L, L))
+        config = np.multiply(config, random_scaling)
 
     for z in range(L):
         for y in range(L):
@@ -70,7 +73,9 @@ def E_dimensionless(config, L, spin_function):
     return total_energy, atomic_features
 
 
-def create_dataset(L, histogram_cutoff, dir, spin_function=lambda x: x):
+def create_dataset(
+    L, histogram_cutoff, dir, spin_function=lambda x: x, scale_spin=False
+):
 
     count_config = 0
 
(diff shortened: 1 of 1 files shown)


examples/ising_model/create_configurations.py
```diff
@@ -26,15 +26,19 @@ def write_to_file(total_energy, atomic_features, count_config, dir):
...
 
     for z in range(L):
         for y in range(L):
             for x in range(L):
-                spin[x, y, z] = math.sin(math.pi * config[x, y, z] / 2)
+                spin[x, y, z] = spin_function(config[x, y, z])
+
+    if scale_spin:
+        random_scaling = np.random.random((L ** 3,))
+        config = np.multiply(config, random_scaling)
```
pzhanggit, line 41: The scaling should be done before `spin_function` otherwise not consistent with https://github.com/ORNL/HydraGNN/pull/87?

streeve: Yeah, you're right. Fixed

examples/ising_model/create_configurations.py
```diff
@@ -100,7 +105,9 @@ def create_dataset(L, histogram_cutoff, dir):
             for config in multiset_permutations(primal_configuration):
                 config = np.array(config)
                 config = np.reshape(config, (L, L, L))
-                total_energy, atomic_features = E_dimensionless(config, L)
+                total_energy, atomic_features = E_dimensionless(
+                    config, L, spin_function
```
pzhanggit, line 109:
```suggestion
                    config, L, spin_function, scale_spin
```

examples/ising_model/create_configurations.py
```diff
@@ -26,15 +26,18 @@ def write_to_file(total_energy, atomic_features, count_config, dir):
...
 
 # 3D Ising model
-def E_dimensionless(config, L):
+def E_dimensionless(config, L, spin_function, scale_spin):
     total_energy = 0
-
     spin = np.zeros_like(config)
 
+    if scale_spin:
+        random_scaling = np.random.random((L ** 3,))
```
pzhanggit, line 34:
```suggestion
        random_scaling = np.random.random((L, L, L))
```

examples/ising_model/create_configurations.py
pzhanggit: Did we have a unit test for Ising model? Not sure why the checks did not catch this

examples/ising_model/create_configurations.py
```diff
@@ -26,15 +26,18 @@ def write_to_file(total_energy, atomic_features, count_config, dir):
...
+        config = np.multiply(config, random_scaling)
+
     for z in range(L):
         for y in range(L):
             for x in range(L):
-                spin[x, y, z] = math.sin(math.pi * config[x, y, z] / 2)
+                spin[x, y, z] = spin_function(config[x, y, z])
```
allaffa, line 40: @streeve 
smart :D

examples/ising_model/create_configurations.py
allaffa: there are currently no uni tests that use this model.

examples/ising_model/create_configurations.py
allaffa: @pzhanggit 
if you propose this change of dimensionality, just make sure that this change of dimensionality does not break the np.multiply in the following line that perform element-wise multiplication between lumpy arrays

pzhanggit: @allaffa  That's exactly why I proposed the change. The original code failed with dimension inconsistency error.

allaffa (approved):

pzhanggit (approved):

PR #94, opened 2022-01-25
Periodic boundary conditions

allaffa:
The Atomic System Environment (ASE) 
https://wiki.fysik.dtu.dk/ase/index.html 
is used to impose periodic boundary conditions by adopting the minimum image convention 
https://en.wikipedia.org/wiki/Periodic_boundary_conditions#Practical_implementation:_continuity_and_the_minimum_image_convention.


hydragnn/preprocess/utils.py
```diff
@@ -20,3 +20,103 @@ def check_if_graph_size_constant(train_loader, val_loader, test_loader):
...
+            assert hasattr(
+                data, "atom_types"
+            ), "The data must contain information about the atoms at each location of the lattice. Can be a chemical symbol (str) or an atomic number (int)."
+            ase_atom_object = ase.Atoms(
+                symbols=data.atom_types,
+                positions=data.pos,
+                cell=data.unit_cell,
+                pbc=True,
+            )
+            edge_src, edge_dst, edge_shift = ase.neighborlist.neighbor_list(
```
streeve, line 78: Do we plan to use the shifts somewhere specific? Maybe I'm misunderstanding how they return the distance info

hydragnn/preprocess/utils.py
```diff
@@ -20,3 +20,103 @@ def check_if_graph_size_constant(train_loader, val_loader, test_loader):
...
+                "ijS", a=ase_atom_object, cutoff=self.r, self_interaction=self.loop
```
streeve, line 79: Can you add a comment on the "ijS"? I think I understand from the docs, but it's kind of a weird input

hydragnn/preprocess/utils.py
```diff
@@ -20,3 +20,103 @@ def check_if_graph_size_constant(train_loader, val_loader, test_loader):
...
                 return graph_size_variable
     return graph_size_variable
+
+
+import ase
+import torch
+from torch_geometric.transforms import RadiusGraph, Distance
+
+
+class RadiusGraphPBC(RadiusGraph):
```
streeve, line 30: I really like this design

hydragnn/preprocess/utils.py
```diff
@@ -20,3 +20,103 @@ def check_if_graph_size_constant(train_loader, val_loader, test_loader):
...
+        return data
+
+    def __repr__(self) -> str:
+        return f"{self.__class__.__name__}(r={self.r})"
+
+
+class DistancePBC(Distance):
+    r"""Saves the Euclidean distance of linked nodes in its edge attributes.
+
+    Args:
```
streeve, line 105: Do we not need to explicitly call the base constructor? Python is crazy

hydragnn/preprocess/utils.py
```diff
@@ -20,3 +20,103 @@ def check_if_graph_size_constant(train_loader, val_loader, test_loader):
...
+            )
+            data.edge_shift = torch.tensor(edge_shift)
+            data.edge_attr = torch.zeros(
+                ase_atom_object.get_global_number_of_atoms(), 1
+            )
+            for index in range(0, edge_src.shape[0]):
+                data.edge_attr[index, 1] = distance_matrix[
+                    edge_src[index], edge_dst[index]
+                ]
+        else:
```
streeve, line 93: Can we just move this else block outside the class (into SerializedLoader)?
```
if config.pbc: return RadiusGraphPBC
else: return RadiusGraph
```
Then we don't need to overload the constructor or store whether we're using PBC

streeve (changes requested):
Way simpler than what I did and a useful library for other details too!

examples/lsms/lsms.json
```diff
@@ -23,6 +23,7 @@
             "model_type": "PNA",
             "radius": 7,
             "max_neighbours": 100000,
+            "periodic_boundary_conditions": false,
```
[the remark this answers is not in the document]
allaffa: For now, I would like to make a case for HydraGNN to be a surrogate for atomic modeling in general.
For some molecular dynamics application I do not think periodic boundary conditions (PBCs) make sense. I do not know what sense it would make to periodically replicate the same molecule over and over again. 
As I see it, PBCs make sense for alloys, but that is only one of the many families that fall under the "atomic modeling" umbrella.

hydragnn/preprocess/utils.py
allaffa: The shift is actually non used for now.

hydragnn/preprocess/utils.py
allaffa: Thanks :D

hydragnn/preprocess/utils.py
allaffa: If I understand correctly how Python works, one needs to explicitly call the base constructor if the child class has a constructor that can be seen as a "specialization" of the base class constructor. 
In this case, the specialized constructor for the child class would be exactly the same as the base class. In fact, I do not re-define the constructor at all. It's hidden from you :D 
But I think you are right. we should make sure that not explicitly calling the base constructor still ensures that everything is preserved from the base class as expected.

hydragnn/preprocess/utils.py
allaffa: I thought about that indeed :)
If we move this if-then-else in the SerializedLoader, we will need to re-create the ASE atomic environment a second time to compute the edge attributes (if needed). 
If one applies PBC, one needs also to make sure that the edge lengths are computed correctly too. To apply both PBC and compute edge lengths, the ASE object is needed. So why not constructing the ASE object and extract everything needed out of it all at once?

examples/lsms/lsms.json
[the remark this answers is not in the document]
streeve: I think you mean "molecular application" and for that I completely agree. But this is the FePt example, which would likely be improved with PBC

hydragnn/preprocess/utils.py
streeve: Can't we just check whether `edge_attr` exist before calling `Distance` in a similar way? I agree that it doesn't make sense to call multiple distance functions. Then we don't even need `DistancePBC`

allaffa: `Distance` creates `edge_attr`. Before calling `Distance`, `edge_attr` does not exist. Moreover, `edge_attr` may contain attributes that are not related to the length of the bond. In that case, `edge_attr` already exists but distance needs to be called anyway to update the existing `edge_attr`

hydragnn/preprocess/utils.py
allaffa: @streeve I removed it :)

hydragnn/preprocess/utils.py
allaffa: Comment added

examples/lsms/lsms.json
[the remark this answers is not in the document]
allaffa: Oh sorry. I thought you meant in general. 
I tried to play with PBC on the FePt. Unfortunately, we cannot. Because three years ago I generated the FePt data using 32 atoms. This system does not equally replicate the unit cell in all the x-y-z directions. This causes the ASE package to mess up the calculation of the PBC. 
In order for PBC to be calculated correctly, the system must replicate the unit cell in all x-y-z directions the same number of times. But I did not think about this three years ago. 

If I knew then what I know now :)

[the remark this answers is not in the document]
streeve: We should just be able to pass a non-uniform cell to ASE - there's no requirement that PBC use uniform cells

hydragnn/preprocess/utils.py
```diff
@@ -20,3 +20,106 @@ def check_if_graph_size_constant(train_loader, val_loader, test_loader):
...
+            # ‘j’ : second atom index
+            # https://wiki.fysik.dtu.dk/ase/ase/neighborlist.html#ase.neighborlist.neighbor_list
+            edge_src, edge_dst = ase.neighborlist.neighbor_list(
+                "ij", a=ase_atom_object, cutoff=self.r, self_interaction=self.loop
+            )
+            distance_matrix = ase_atom_object.get_all_distances(mic=True)
+            data.edge_index = torch.stack(
+                [torch.LongTensor(edge_src), torch.LongTensor(edge_dst)], dim=0
+            )
+            data.edge_shift = torch.tensor(edge_shift)
```
streeve, line 88: Remove :)

examples/lsms/lsms.json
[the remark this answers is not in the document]
allaffa: Can you define what non-uniform cell is?

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -63,8 +67,12 @@ def load_serialized_data(self, dataset_path: str, config):
...
-        compute_edge_lengths = Distance(norm=False, cat=True)
+        if config["Architecture"]["periodic_boundary_conditions"]:
+            # edge lengths already added manually if using PBC, so no need to call Distance.
+            compute_edges = get_radius_graph_pbc_config(config["Architecture"])
+        else:
+            compute_edges = get_radius_graph_config(config["Architecture"])
+            compute_edge_lengths = Distance(norm=False, cat=True)
 
         dataset[:] = [compute_edges(data) for data in dataset]
         dataset[:] = [compute_edge_lengths(data) for data in dataset]
```
allaffa, line 78: @streeve 
Differently from RadiusGraph the computes only the adjacency matrix WITHOUT computing the length of the edges, RadiusGraphPBC computes both the adjacency matrix AND the length of the edges. Therefore, once RadiusGraphPBC is called,  you do not need to call compute_edge_lengths anymore. 

Your changes are not valid

streeve: Fixed, I think

examples/lsms/lsms.json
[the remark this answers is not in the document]
streeve: Because we have fixed angles, I mean a length 3 vector for the xyz lattice size instead of a single lattice parameter. The fully general case would be a 3x3 matrix with variable lengths and angles in xyz 
https://wiki.fysik.dtu.dk/ase/ase/atoms.html

hydragnn/preprocess/serialized_dataset_loader.py
streeve: Discussed in call

tests/inputs/ci.json
```diff
@@ -26,6 +26,7 @@
             "model_type": "PNA",
             "radius": 2.0,
             "max_neighbours": 100,
+            "periodic_boundary_conditions": false,
```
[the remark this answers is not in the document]
allaffa: Ising model?

streeve:
Finally pushed a state that should work for now. As we discussed, we now have a test where there should be no periodic neighbors (isolated molecule) and one where every node has equal neighbors once PBC are applied. This does not handle cases where the system size is less than half the cutoff (quite common in DFT), but we can address that in the future

allaffa:
> Finally pushed a state that should work for now. As we discussed, we now have a test where there should be no periodic neighbors (isolated molecule) and one where every node has equal neighbors once PBC are applied. This does not handle cases where the system size is less than half the cutoff (quite common in DFT), but we can address that in the future

If unit tests pass, approve the PR. Since I created the PR, I cannot approve it myself.

hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -65,14 +69,26 @@ def load_serialized_data(self, dataset_path: str, config):
...
+                config["NeuralNetwork"]["Architecture"]
+            )
+            compute_edge_lengths = Distance(norm=False, cat=True)
+
         dataset[:] = [compute_edges(data) for data in dataset]
-        dataset[:] = [compute_edge_lengths(data) for data in dataset]
+
+        # edge lengths already added manually if using PBC.
+        if not config["NeuralNetwork"]["Architecture"]["periodic_boundary_conditions"]:
+            compute_edge_lengths = Distance(norm=False, cat=True)
```
pzhanggit, line 90: repeated with line 84?

streeve: Thanks, fixed

streeve (approved):

streeve:
I squashed some commits together, but left the most recent few to keep how you dealt with duplicates. I approved, but would prefer to wait until Pei also does

PR #95, opened 2022-01-25
Eam data loader

allaffa:


hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -181,7 +181,7 @@ def update_predicted_values(type: list, index: list, data: Data):
         A Data object representing a structure that has atoms.
     """
     output_feature = []
-    data.y_loc = torch.zeros(1, len(type) + 1, dtype=torch.int64, device=data.y.device)
+    data.y_loc = torch.zeros(1, len(type) + 1, dtype=torch.int64, device=data.x.device)
```
streeve, line 184: Can you explain this change?

hydragnn/preprocess/raw_dataset_loader.py
```diff
@@ -91,15 +98,24 @@ def load_raw_data(self):
...
                 all_lines = f.readlines()
-                data_object = self.__transform_input_to_data_object_base(
-                    lines=all_lines
-                )
+
+                if self.data_format == "LSMS":
+                    data_object = self.__transform_LSMS_input_to_data_object_base(
+                        lines=all_lines
+                    )
+                elif self.data_format == "EAM":
```
streeve, line 106: If I understand correctly, the data you're reading here is CFG - EAM is the physical model being used

streeve (changes requested):

hydragnn/preprocess/raw_dataset_loader.py
```diff
@@ -91,15 +98,24 @@ def load_raw_data(self):
...
                 f.close()
 
-            if self.data_format == "LSMS":
-                for idx, data_object in enumerate(dataset):
-                    dataset[idx] = self.__charge_density_update_for_LSMS(data_object)
+                if self.data_format == "LSMS":
+                    for idx, data_object in enumerate(dataset):
+                        dataset[idx] = self.__charge_density_update_for_LSMS(
+                            data_object
+                        )
```
jychoi-hpc, line 118: I found this block is getting slowing down as it progresses. Basically, this block is a nested block. Is this intentional? Is it ok to move up this block to put this outside of for-loop (line 97)?

allaffa: Yes, you are right. Essentially we would just need to un-indentation this block. Correct?

jychoi-hpc: Yes, it is correct.

hydragnn/preprocess/serialized_dataset_loader.py
allaffa: data.y is not defined if there are NO global features in the data.
data.x is always defined.

examples/eam/eam_multitask.json
```diff
@@ -0,0 +1,62 @@
...
+    },
+    "Dataset": {
+        "name": "NiNb",
+        "path": {"total": "FCC_Ni_Nb"},
+        "format": "CFG",
+        "compositional_stratified_splitting": true,
+	"rotational_invariance": true,
+        "node_features": {
+            "name": ["num_of_protons", "atomic_mass", "atomic_energy", "atomic_force"],
+            "dim": [1,1,1,3],
```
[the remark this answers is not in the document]
allaffa: If graph properties are vectors, yes it is.

PR #96, opened 2022-01-29
Logging feature

jychoi-hpc:
Add a python logging feature (`https://docs.python.org/3/library/logging.html`) to print messages to the screen as well as to a file (`run.log`).

Three functions are added:
* `logging_init`: setup logging
* `info`: print message prefixed with "[rank:INFO]"
* `log`: print message prefixed with "[rank:DEBUG]"

### DIFF
--- a/hydragnn/run_training.py
+++ b/hydragnn/run_training.py
@@ -58,8 +58,8 @@ def _(config: dict):
     except:
         os.environ["SERIALIZED_DATA_PATH"] = os.getcwd()
 
+    setup_log(get_log_name_config(config))
     world_size, world_rank = setup_ddp()
-    setup_log(get_log_name_config(config), world_size)
 
     verbosity = config["Verbosity"]["level"]
     train_loader, val_loader, test_loader = dataset_loading_and_splitting(config=config)
--- a/hydragnn/run_training.py
+++ b/hydragnn/run_training.py
@@ -26,7 +26,7 @@ from hydragnn.utils.model import (
     get_summary_writer,
     load_existing_model_config,
 )
-from hydragnn.utils.print_utils import print_distributed, logging_init
+from hydragnn.utils.print_utils import print_distributed, setup_log
 from hydragnn.utils.time_utils import print_timers
 from hydragnn.utils.config_utils import (
     update_config,
@@ -59,9 +59,7 @@ def _(config: dict):
         os.environ["SERIALIZED_DATA_PATH"] = os.getcwd()
 
     world_size, world_rank = setup_ddp()
-
-    prefix = get_log_name_config(config)
-    logging_init(prefix, world_rank)
+    setup_log(get_log_name_config(config), world_size)
 
     verbosity = config["Verbosity"]["level"]
     train_loader, val_loader, test_loader = dataset_loading_and_splitting(config=config)
--- a/hydragnn/run_training.py
+++ b/hydragnn/run_training.py
@@ -26,7 +26,7 @@ from hydragnn.utils.model import (
     get_summary_writer,
     load_existing_model_config,
 )
-from hydragnn.utils.print_utils import print_distributed, logging_init, log, info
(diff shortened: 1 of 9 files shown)


hydragnn/utils/print_utils.py
```diff
@@ -63,3 +67,42 @@ def iterate_tqdm(iterator, verbosity_level):
...
+    handlers.append(logging.FileHandler(fname))
+    logging.basicConfig(level=logging.DEBUG, format=fmt, handlers=handlers)
+
+
+"""
+uage: info (str1, str2, ...). Use just like print. "[rank:INFO]" will be prefixed.
+"""
+
+
+def info(*args, logtype="info", sep=" "):
```
streeve, line 94: Not sure I follow the difference between info and log

streeve (commented):
Looks very nice! Looks like real errors in the CI though

hydragnn/utils/print_utils.py
jychoi-hpc: The difference is logging level. `info()` will tag outputs with "INFO" and log will tag outputs with "DEBUG". Later we can make an option to suppress "DEBUG" messages for cleaner output.

streeve: How about `log()` "LOG" and `debug()` "DEBUG". Unless what you're suggesting is standard

hydragnn/run_training.py
```diff
@@ -60,6 +60,9 @@ def _(config: dict):
 
     world_size, world_rank = setup_ddp()
 
+    prefix = get_log_name_config(config)
+    logging_init(prefix, world_rank)
```
streeve, line 64: `setup_log` to match `setup_ddp` a little more?

jychoi-hpc: Yes it makes sense. I have changed.

hydragnn/utils/print_utils.py
jychoi-hpc: I have changed too. `into()` to "LOG()", `debug()` to "DEBUG()"

streeve:
I read more about the `logging` package and understand a little better. Now I see why you named them `info` and `debug`, but I'm hoping we can integrate this with `print_distributed` a little more. I would like to able to print debug, general info, and warnings (maybe errors?) with a given verbosity throughout the code - do you see a clean way to do so? If not, maybe we just use the info level wrapped inside as you have now

streeve:
That sounds great to me

hydragnn/utils/print_utils.py
```diff
@@ -63,3 +67,39 @@ def iterate_tqdm(iterator, verbosity_level):
         return tqdm(iterator)
     else:
         return iterator
+
+
+"""
+Setup logging to print messages for both screen and file.
```
streeve, line 73: I think if you move the docstrings in the functions `black` will stop forcing these extra blank lines

hydragnn/utils/print_utils.py
streeve: Sorry for the confusion I caused. Like you say below, functions for `debug()`, `info()`, and `warning()` work for me

jychoi-hpc:
I found adding the concept of log level (debug, info, warning) needs more discussion among us (needs some changes in many places). Instead, I figured out a clean way without breaking the current code.

The logging is same as before. Now, we can print both screen and save to `run.log` at the same time.

streeve (approved):
I like this a lot - it does exactly what we need. I tested locally and added to the examples as well

streeve:
I okay to squash merge this once someone else approves

allaffa (approved):

streeve:
Ready to merge after fixing issues with python 3.7

jychoi-hpc:
Made modification to work with python 3.7 too. I tested this working with python 3..7, 3.8, and 3.9.

pzhanggit (approved):

PR #97, opened 2022-01-29
Compositional histogram cutoff 

allaffa:


utils/compositional_histogram_cutoff.py
```diff
@@ -0,0 +1,45 @@
...
+
+    element_counter = dict()
+    composition_counter = dict()
+
+    for filename in os.listdir(path_to_dir):
+
+        for atom_type in elements_list:
+            element_counter[atom_type] = 0
+
+        df = pandas.read_csv(path_to_dir + filename, header=None, skiprows=1)
```
streeve, line 25: Can you steal from the previous utils PR to avoid the pandas dependency. I can update if needed

allaffa: Steal as much as you want :-)

streeve (approved):

streeve:
This is ready for review if you want to take a look @pzhanggit. I approved, but I also wrote part of it

utils/lsms/convert_total_energy_to_formation_gibbs.py
```diff
@@ -27,7 +28,7 @@ def read_file(path):
 
 
 def convert_raw_data_energy_to_gibbs(
-    dir, elements_list, temperature_kelvin=0, create_plots=False
+    dir, elements_list, temperature_kelvin=0, overwrite_data=False, create_plots=True
```
pzhanggit, line 31: It seems `overwrite_data=False` would not work and the data files will be overwritten anyway?

utils/lsms/compositional_histogram_cutoff.py
```diff
@@ -0,0 +1,63 @@
...
+    comp_all = np.zeros([num_bins])
+    for filename in tqdm(os.listdir(dir)):
+
+        path = os.path.join(dir, filename)
+        # This is LSMS specific - it assumes only one header line and only atoms following.
+        atoms = np.loadtxt(path, skiprows=1)
+
+        elements, counts = np.unique(atoms[:, 0], return_counts=True)
+        num_atoms = atoms.shape[0]
+        # This will result in composition of 1.0 for any pure element.
```
pzhanggit, line 45: Is it going to be a problem that any pure element will be categorized to the same bin and hence use the same histogram_cutoff?

pzhanggit (commented):
Is the function for systems with two elements only?  It the function currently called anywhere?
It looks good to me overall. I only have a few questions.

allaffa:
> Is the function for systems with two elements only? It the function currently called anywhere?
> It looks good to me overall. I only have a few questions.

@pzhanggit 

yes, for now the implementation works only for binaries. I would be more than happy to answer your questions before merging the PR.

utils/lsms/convert_total_energy_to_formation_gibbs.py
streeve: Fixed

PR #100, opened 2022-02-07
Add initial PI3NN example

streeve:
Add general functions for calculating confidence intervals with PI3NN and an example for the existing FePt case.

Things to discuss
- There's some cleanup I need to do where I modified the main examples/training files
- Should any of the UQ functions move to the library? Maybe up/down loaders?
- PI3NN is not currently installed/used in the CI. I'd prefer to wait until tensorflow is optional (not used in the very small part of the library we're calling)
- This only currently works for single task - should not be too difficult to extend


examples/lsms/lsms.json
```diff
@@ -4,7 +4,7 @@
     },
     "Dataset": {
         "name": "FePt_32atoms",
-        "path": {"total": "FePt"},
+        "path": {"total": "../../Downloads/FePt"},
```
allaffa, line 7: Is this intentional?

examples/lsms/lsms.json
```diff
@@ -38,17 +38,17 @@
...
-            "task_weights": [1.0, 1.0, 1.0]
+            "task_weights": [1.0]
         },
         "Variables_of_interest": {
             "input_node_features": [0],
-            "output_names": ["free_energy_scaled_num_nodes","charge_density", "magnetic_moment"],
-            "output_index": [0, 1, 2],
-            "type": ["graph","node","node"],
-            "denormalize_output": true
+            "output_names": ["free_energy_scaled_num_nodes"],
```
allaffa, line 45: I would not remove the multi-task nature of the run based on `lsms.json`

examples/lsms_uq.json
```diff
@@ -0,0 +1,57 @@
+{
```
allaffa, line 1: Given indeed that you are creating another .json file specifically for UQ, I do not understand why you changed `lsms.json`

hydragnn/run_training.py
```diff
@@ -76,12 +76,12 @@ def _(config: dict):
         optimizer, mode="min", factor=0.5, patience=5, min_lr=0.00001
     )
 
-    log_name = get_log_name_config(config)
-    writer = get_summary_writer(log_name)
+    output_name = model.__str__() + "-mean"
```
allaffa, line 79: Why did you remove the call to `get_log_name_config`?
Is it needed specifically for UQ or is it a proposed improvement that applies in general to any training?

examples/lsms/lsms.json
allaffa: I would not change the original `lsms.json`

streeve: No, like I said in the description, I realized after creating the PR that I need to remove some changes that weren't meant to be included

hydragnn/run_training.py
streeve: I think adding an option to override the name would be nice, but I will leave the default behavior intact

examples/uq/uq_pi3nn.py
```diff
@@ -0,0 +1,342 @@
...
+    Create the up and down datasets by splitting on mean model predictions.
+    """
+    device = next(model.parameters()).device
+    verbosity = config["Verbosity"]["level"]
+    batch_size = config["NeuralNetwork"]["Training"]["batch_size"]
+
+    model.eval()
+
+    rank = 0
+    if dist.is_initialized():
```
allaffa, line 249: Is the current version tested on MPI ?

streeve: Good point - no it is not

streeve:
| Do you need additional review for this?

I cleaned up the issues I saw in the diff - another review would be helpful

examples/uq/uq_pi3nn.py
```diff
@@ -0,0 +1,344 @@
...
+    try:
+        os.environ["SERIALIZED_DATA_PATH"]
+    except:
+        os.environ["SERIALIZED_DATA_PATH"] = os.getcwd()
+
+    world_size, world_rank = setup_ddp()
+    verbosity = config["Verbosity"]["level"]
+
+    #### LOAD THE ORIGINAL DATA LOADERS AND THE TRAINED MEAN MODEL
+    ## Question: Should we normalize/denormalize any input/output?
```
pzhanggit, line 71: Maybe normalize everything before the three NNs and denormalize them after all of them trained

examples/uq/uq_pi3nn.py
```diff
@@ -0,0 +1,344 @@
...
+        yx.squeeze() - 0.001,
+        (pred_mean_y - c_down * pred_down_y),
+        edgecolor="#000066",
+        marker="o",
+        facecolor="none",
+    )
+    plt.show()
+
+
+def load_model(config, dataset, name):
```
pzhanggit, line 177: `dataset` argument is not needed?

examples/uq/uq_pi3nn.py
```diff
@@ -0,0 +1,344 @@
...
+                down, batch_size=batch_size, shuffle=False, sampler=down_sampler
+            )
+        else:
+            up_loader = DataLoader(up, batch_size=batch_size, shuffle=True)
+            down_loader = DataLoader(down, batch_size=batch_size, shuffle=True)
+
+        up_loaders.append(up_loader)
+        down_loaders.append(down_loader)
+
+    return up_loaders, down_loaders
```
pzhanggit, line 306: `up_sampler` and `down_sampler` need to be returned also since they are needed for shuffling during training in each epoch, see here https://github.com/streeve/HydraGNN/blob/2e93dc50d1744fd5a2e5eaef9f4531ee2ff8093a/hydragnn/train/train_validate_test.py#L94

pzhanggit (changes requested):
Finish first round review. I will come back later for more reviews

examples/uq/uq_pi3nn.py
```diff
@@ -0,0 +1,344 @@
...
+
+def run_uncertainty(
+    config_file_mean, config_file_up_down, retrain_mean=False, retrain_up_down=False
+):
+    """
+    Compute prediction intervals with PI3NN.
+    """
+
+    out_name = "uq_"
+    mean_name = out_name + "mean"
```
streeve, line 54: mean -> center

streeve, line 54: mean -> center

PR #102, opened 2022-02-15
rotational invariance added

allaffa:
I am using the PyTorch Geometric transformation called `NormalizeRotation` 
https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/transforms/normalize_rotation.html#NormalizeRotation

This transformation computes the eigenvectors if the point could, and transforms data.pos by applying a rotation with respect to the eigenvectors. 

This transformation ensures that data.pos is rotational invariant. Any rotation applied to the graph affects the eigenvectors of the point cloud, but **the relative coordinates of data.pos with respect to the eigenvectors DO NOT CHANGE**

Part of #2

### DIFF
--- a/hydragnn/preprocess/serialized_dataset_loader.py
+++ b/hydragnn/preprocess/serialized_dataset_loader.py
@@ -15,9 +15,10 @@ from sklearn.model_selection import StratifiedShuffleSplit
 
 import torch
 from torch_geometric.data import Data
-from torch_geometric.transforms import RadiusGraph, Distance, NormalizeRotation
+from torch_geometric.transforms import Distance, NormalizeRotation
 
 from .dataset_descriptors import AtomFeatures
+from hydragnn.preprocess import get_radius_graph_config
 from hydragnn.utils.distributed import get_device
 from hydragnn.utils.print_utils import print_distributed, iterate_tqdm
 
@@ -64,7 +65,7 @@ class SerializedDataLoader:
             dataset = pickle.load(f)
 
         rotational_invariance = NormalizeRotation(max_points=-1, sort=False)
-        compute_edges = get_radius_graph(config["NeuralNetwork"]["Architecture"])
+        compute_edges = get_radius_graph_config(config["NeuralNetwork"]["Architecture"])
         compute_edge_lengths = Distance(norm=False, cat=True)
 
         if config["Dataset"]["rotational_invariance"]:
@@ -169,14 +170,6 @@ class SerializedDataLoader:
             subsample.append(dataset[index])
 
         return subsample
-
-
-def get_radius_graph(config):
-    return RadiusGraph(
-        r=config["radius"],
-        loop=False,
-        max_num_neighbors=config["max_neighbours"],
-    )
 
 
 def update_predicted_values(type: list, index: list, data: Data):
--- a/hydragnn/preprocess/serialized_dataset_loader.py
+++ b/hydragnn/preprocess/serialized_dataset_loader.py
@@ -15,7 +15,7 @@ from sklearn.model_selection import StratifiedShuffleSplit
 
 import torch
 from torch_geometric.data import Data
-from torch_geometric.transforms import RadiusGraph, Distance
+from torch_geometric.transforms import RadiusGraph, Distance, NormalizeRotation
 
 from .dataset_descriptors import AtomFeatures
 from hydragnn.utils.distributed import get_device
@@ -63,8 +63,12 @@ class SerializedDataLoader:
             _ = pickle.load(f)
             dataset = pickle.load(f)
 
+        rotational_invariance = NormalizeRotation(max_points=-1, sort=False)
         compute_edges = get_radius_graph(config["NeuralNetwork"]["Architecture"])
         compute_edge_lengths = Distance(norm=False, cat=True)
+
+        if config["Dataset"]["rotational_invariance"]:
+            dataset[:] = [rotational_invariance(data) for data in dataset]
 
         dataset[:] = [compute_edges(data) for data in dataset]
         dataset[:] = [compute_edge_lengths(data) for data in dataset]
--- a/tests/test_rotational_invariance.py
+++ b/tests/test_rotational_invariance.py
@@ -1,0 +1,127 @@
+##############################################################################
+# Copyright (c) 2021, Oak Ridge National Laboratory                          #
+# All rights reserved.                                                       #
+#                                                                            #
+# This file is part of HydraGNN and is distributed under a BSD 3-clause      #
+# license. For the licensing terms see the LICENSE file in the top-level     #
+# directory.                                                                 #
+#                                                                            #
+# SPDX-License-Identifier: BSD-3-Clause                                      #
+##############################################################################
+
+import sys, os, json
+import pytest
+
+import torch
+from torch_geometric.data import Data
+from torch_geometric.transforms import Distance, NormalizeRotation
+from hydragnn.preprocess.utils import (
+    get_radius_graph_config,
+)
+
+from hydragnn.preprocess.utils import (
+    check_data_samples_equivalence,
+)
+
+torch.set_default_tensor_type(torch.DoubleTensor)
+
+
+def unittest_rotational_invariance():
+    config_file = "./tests/inputs/ci_rotational_invariance.json"
+    config = {}
+    with open(config_file, "r") as f:
+        config = json.load(f)
+    compute_edges = get_radius_graph_config(config["Architecture"], loop=False)
+    compute_edge_lengths = Distance(norm=False, cat=True)
+    rotation = NormalizeRotation(max_points=-1, sort=False)
+
+    # Create BCT lattice structure with 32 nodes
+    data = Data()
+
+    data.pos = torch.tensor(
+        [
+            [0.000000000000000, 0.000000000000000, 0.000000000000000],
+            [2.609000000000000, 2.609000000000000, 3.529000000000000],
+            [5.218000000000000, 0.000000000000000, 0.000000000000000],
+            [7.827000000000000, 2.609000000000000, 3.529000000000000],
+            [0.000000000000000, 5.218000000000000, 0.000000000000000],
+            [2.609000000000000, 7.827000000000000, 3.529000000000000],
+            [5.218000000000000, 5.218000000000000, 0.000000000000000],
+            [7.827000000000000, 7.827000000000000, 3.529000000000000],
+            [0.000000000000000, 0.000000000000000, 7.058000000000000],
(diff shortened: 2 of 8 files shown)


hydragnn/preprocess/serialized_dataset_loader.py
```diff
@@ -63,9 +63,11 @@ def load_serialized_data(self, dataset_path: str, config):
             _ = pickle.load(f)
             dataset = pickle.load(f)
 
+        rotational_invariance = NormalizeRotation(max_points=-1, sort=False)
```
streeve, line 66: Should this be optional?

streeve (commented):
This absolutely makes sense - have you seen  accuracy changes on any dataset?

hydragnn/preprocess/serialized_dataset_loader.py
allaffa: Good question. For now, I think all we do I is predicting properties that are rotational invariant. 
But you never know what happens in the future. In situations where we may need to use it for surface interaction, maybe rotational invariance is not good? I don't know...

allaffa:
> This absolutely makes sense - have you seen accuracy changes on any dataset?

@streeve 
The unit tests on graphs defined nodal and global properties that are strongly dependent on x-y-z coordinates. Rotating the graph impacts x-y-z coordinates, which in turn impacts the value of global and nodal properties. 

Should I change the existing unit tests or should I create new ones?
Maybe if we make rotation-invariance optional is better to make a new set of unit tests.

pzhanggit:
> > This absolutely makes sense - have you seen accuracy changes on any dataset?
> 
> @streeve The unit tests on graphs defined nodal and global properties that are strongly dependent on x-y-z coordinates. Rotating the graph impacts x-y-z coordinates, which in turn impacts the value of global and nodal properties.
> 
> Should I change the existing unit tests or should I create new ones? Maybe if we make rotation-invariance optional is better to make a new set of unit tests.

That's interesting. Did you see any change in accuracy? My expectation (correct me if I am wrong) is that the changes made in this PR will not change the unit test.  The nodal and global properties are generated before this function and hence would not be affected. In addition, the rotation preserves the relative position between nodes and hence would not affect the edge construction either.

allaffa:
> > > This absolutely makes sense - have you seen accuracy changes on any dataset?
> > 
> > 
> > @streeve The unit tests on graphs defined nodal and global properties that are strongly dependent on x-y-z coordinates. Rotating the graph impacts x-y-z coordinates, which in turn impacts the value of global and nodal properties.
> > Should I change the existing unit tests or should I create new ones? Maybe if we make rotation-invariance optional is better to make a new set of unit tests.
> 
> That's interesting. Did you see any change in accuracy? My expectation (correct me if I am wrong) is that the changes made in this PR will not change the unit test. The nodal and global properties are generated before this function and hence would not be affected. In addition, the rotation preserves the relative position between nodes and hence would not affect the edge construction either.

@pzhanggit we explicitly use the absolute values of x-y-z coordinates to compute global properties and nodal properties of the graph. Therefore, the definition of the properties depends on the coordinate system used to define the x-y-z coordinates. Right?

allaffa:
> > > > > This absolutely makes sense - have you seen accuracy changes on any dataset?
> > > > 
> > > > 
> > > > @streeve The unit tests on graphs defined nodal and global properties that are strongly dependent on x-y-z coordinates. Rotating the graph impacts x-y-z coordinates, which in turn impacts the value of global and nodal properties.
> > > > Should I change the existing unit tests or should I create new ones? Maybe if we make rotation-invariance optional is better to make a new set of unit tests.
> > > 
> > > 
> > > That's interesting. Did you see any change in accuracy? My expectation (correct me if I am wrong) is that the changes made in this PR will not change the unit test. The nodal and global properties are generated before this function and hence would not be affected. In addition, the rotation preserves the relative position between nodes and hence would not affect the edge construction either.
> > 
> > 
> > @pzhanggit we explicitly use the absolute values of x-y-z coordinates to compute global properties and nodal properties of the graph. Therefore, the definition of the properties depends on the coordinate system used to define the x-y-z coordinates. Right?
> 
> Yes, but the properties are defined in `deterministic_graph_data` before "NormalizeRotation" is called to change x-y-z.

Niiiiiiiiiiice :D

allaffa:
@pzhanggit @streeve 

Can you explain why you would expect improvement in accuracy?

pzhanggit:
> @pzhanggit @streeve
> 
> Can you explain why you would expect improvement in accuracy?

I am not expecting improvement in accuracy. More specific, I am actually expecting this PR would change nothing in the unit tests or the other cases we have.

streeve:
> Can you explain why you would expect improvement in accuracy?

For the unit tests, I agree: we don't rotate anything. For QM9 I _would_ expect improvement if it helps the network understand that orientation of any given molecule doesn't matter

allaffa:
> > Can you explain why you would expect improvement in accuracy?
> 
> For the unit tests, I agree: we don't rotate anything. For QM9 I _would_ expect improvement if it helps the network understand that orientation of any given molecule doesn't matter

**The rotational invariance improves prediction on QM9 (or any other dataset) ONLY IF angle features are used as input features. In the current version of the code, angle features are not used**. 
Therefore, I do not see why the rotational invariance alone should improve the predictions on QM9.
If you still think results should improve, please explain.

streeve:
Our discussion in person yesterday resolves this discussion - I agree with you. I think that making it optional is important based on the fact that it is not generally useful

allaffa:
> Our discussion in person yesterday resolves this discussion - I agree with you. I think that making it optional is important based on the fact that it is not generally useful

Thank you! I had committed the changes where I make it optional, but I had forgotten to push. 
Optional flag currently used ;)

tests/test_graphs.py
```diff
@@ -178,10 +183,25 @@ def unittest_train_model(model_type, ci_input, use_lengths, overwrite_data=False
...
 @pytest.mark.parametrize("ci_input", ["ci.json", "ci_multihead.json"])
 def pytest_train_model(model_type, ci_input, overwrite_data=False):
-    unittest_train_model(model_type, ci_input, False, overwrite_data)
+    unittest_train_model(model_type, ci_input, False, False, overwrite_data)
+
+
+# Test across all models with both single/multihead
+@pytest.mark.parametrize("model_type", ["GIN", "GAT", "MFC", "PNA", "CGCNN"])
+@pytest.mark.parametrize("ci_input", ["ci.json", "ci_multihead.json"])
+def pytest_train_model_rotational_invariance(
```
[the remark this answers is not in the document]
allaffa: We don't expect changes, but we also don't expect the code to behave worse. I thought this would be a sanity check that make sure that capabilities built on top of each other are still compatible.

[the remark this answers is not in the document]
streeve: That's a good point - maybe we only run one model instead of all 5 for now then?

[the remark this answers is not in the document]
allaffa: Are bargaining? Let's make 2. PNA and GCNN, which are the two we mostly care of so far. :-)

allaffa:
@streeve 

I reduced the number of models checked with rotation of the data in test_graph.py
Moreover, I created a new test which does the following:

- check consistency before and after rotation of the graph generated from BCT structure associated with FePt
- checks consistency before and after rotation of 10 random graphs

allaffa:
@streeve @pzhanggit 
I need you guys to check the implementation of this unit test
The assert on the edge length seems to fail whenever I set a tolerance below 1e-5. 
It seems that `torch.geometric.transformations.Distance` does quite a poor job in computing the edge length from the x-y-z coordinates. 
I need you to confirm that I did not create bugs in the code. 

If my code is bug-free, and `torch.geometric.transformations.Distance` really does quite a poor job , we need to be careful. 
Forces are the derivative of the energy with respect to the atom's positions and their neighbors. 
This could affect the use of HydraGNN for prediction on forces from energy.

examples/lsms/lsms.json
```diff
@@ -7,6 +7,7 @@
         "path": {"total": "FePt"},
         "format": "LSMS",
         "compositional_stratified_splitting": true,
+        "rotational_invariance": true,
```
streeve, line 10: Should we switch back to false?

tests/test_rotational_invariance.py
```diff
@@ -0,0 +1,144 @@
...
+
+import sys, os, json
+import pytest
+
+import torch
+from torch_geometric.data import Data
+from torch_geometric.transforms import RadiusGraph, Distance, NormalizeRotation
+
+
+def unittest_rotational_invariance():
```
streeve, line 20: This looks right to me - I would split everything after creating the structure (random or crystal) into a helper function for the rotation/asserts

tests/test_rotational_invariance.py
```diff
@@ -0,0 +1,144 @@
...
+
+    compute_edges = RadiusGraph(r=7.0, loop=False)
```
streeve, line 22: We could use our function wrappers here

streeve (changes requested):
I think the accuracy makes sense for single precision

pzhanggit:
> I think the accuracy makes sense for single precision

Agreed. Default torch type is torch.float32

examples/lsms/lsms.json
allaffa: I removed it.

tests/test_rotational_invariance.py
allaffa: The helper function has been created and added to 
`hydrogen.preprocess.util.py`

tests/test_rotational_invariance.py
allaffa: Done.

pzhanggit (commented):
Looks good to me. One thing is that maybe we do not need full unit tests (PNA and CGCNN) with accuracy checking for rotation, given that data object invariance is checked already in`unittest_rotational_invariance()`. One or two epochs like in qm9/md17 for functionality checking would work IMO.

tests/test_rotational_invariance.py
allaffa: Done.

allaffa:
@streeve @pzhanggit 
I am using double precision in the rotational invariance unit test. With double precision, I can lower the tolerance down to 1e-14 and the test still passes. 
With this (more solid and confident( check in place, I removed the additional training of PNA and CGCNN that I was using before. This way, the CI takes approximately the same time as before to run.

streeve:
@allaffa fixed this up - I think it's ready. @pzhanggit can you take a quick look as well?

streeve (approved):

pzhanggit (approved):
Looks good to me.

PR #106, opened 2022-02-26
Set shuffle true for DistributedSampler

jychoi-hpc:
A quick fix to set `shuffle=true` for DistributedSampler.


jychoi-hpc:
Pei and Max, you are right.  `shuffle=true` is default for `DistributedSampler`. I will rollback the changes.

jychoi-hpc:
I will drop this PR since the shuffle setting is already correct in the main repo:
* With distributed training, we need to set shuffle=True for DistributedSampler, shuffle=False for DataLoader
* With a serial training, shuffle=True for DataLoader

PR #108, opened 2022-03-04
Enable user specified backend

### LINKED ISSUE #101: Support MPI backend
This is mostly motivated by the use case of running one instance of Hydra per GPU on a node. I was only able to do so with the `mpi` backend and iterating through the `LSB_HOSTS` list.

@jychoi-hpc if there's a better way, please let me know; otherwise, I'll plan to add something like what I describe

streeve:
Allow the user to choose the backend, still defaulting to `nccl`, and adding `mpi`

I previously ran multiple separate instances on one node with `mpi`, but that relied on the machine mapping each run to each device. We probably want to explicitly do that in `get_device_name`

This also removes checking for the backend being available, allowing torch to error on its own

Closes #101

### DIFF
--- a/hydragnn/run_prediction.py
+++ b/hydragnn/run_prediction.py
@@ -25,28 +25,28 @@ from hydragnn.postprocess.postprocess import output_denormalize
 
 
 @singledispatch
-def run_prediction(config):
+def run_prediction(config, backend=None):
     raise TypeError("Input must be filename string or configuration dictionary.")
 
 
 @run_prediction.register
-def _(config_file: str):
+def _(config_file: str, backend=None):
 
     with open(config_file, "r") as f:
         config = json.load(f)
 
-    run_prediction(config)
+    run_prediction(config, backend)
 
 
 @run_prediction.register
-def _(config: dict):
+def _(config: dict, backend=None):
 
     try:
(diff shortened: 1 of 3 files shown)


pzhanggit:
Could we put `backend` into `config`?

jychoi-hpc:
I am wondering if we can envs to set runtime settings (such as backend) and `config` file for something we don't need to change run by run/system by system. 

Here is something I thought to set backend with envs:
```
    if os.getenv("HYDRAGNN_BACKEND") is not None:
        backend =os.environ["HYDRAGNN_BACKEND"]
    elif dist.is_mpi_available():
        backend = "mpi"
    elif dist.is_nccl_available() and torch.cuda.is_available():
        backend = "nccl"
    elif torch.distributed.is_gloo_available():
        backend = "gloo"
    else:
        raise RuntimeError("No parallel backends available")
```

Some advantages of using env instead of `config` file would be:
* Setting a backend is a one time at the beginning of run. No need to change code in multiple places
* No need to change `config` file for different systems
Cons:
* user need to set ENVS in the job script

This is just my thought.

jychoi-hpc:
I will address this feature in the upcoming OGB branch.

PR #109, opened 2022-03-06
SAGEStack added

allaffa:
This is one of the simplest convolutional layers.
The introduction of this additional convolutional layer is minimally invasive with respect to the existing version of the repository, and helps us move towards the goal of introducing hyper parameter optimization in the repo, where the choice of the convolutional layer (GIN, GAT, PNA, CGCNN, SAGE) will be an hyper parameter to optimize to maximize validation accuracy.

### DIFF
--- a/hydragnn/models/SAGEStack.py
+++ b/hydragnn/models/SAGEStack.py
@@ -19,34 +19,8 @@ from .Base import Base
 
 
 class SAGEStack(Base):
-    def __init__(
-        self,
-        input_dim: int,
-        output_dim: list,
-        output_type: list,
-        num_nodes: int,
-        hidden_dim: int,
-        config_heads: {},
-        dropout: float = 0.25,
-        num_conv_layers: int = 16,
-        ilossweights_hyperp: int = 1,  # if =1, considering weighted losses for different tasks and treat the weights as hyper parameters
-        loss_weights: list = [1.0, 1.0, 1.0],  # weights for losses of different tasks
-        ilossweights_nll: int = 0,  # if =1, using the scalar uncertainty as weights, as in paper
-        # https://openaccess.thecvf.com/content_cvpr_2018/papers/Kendall_Multi-Task_Learning_Using_CVPR_2018_paper.pdf
-    ):
-        super().__init__(
-            input_dim,
-            hidden_dim,
-            output_dim,
-            output_type,
-            config_heads,
-            ilossweights_hyperp,
-            loss_weights,
-            ilossweights_nll,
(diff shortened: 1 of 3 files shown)


streeve:
Waiting on Pei's review? Or is this ready to merge?

allaffa:
> Waiting on Pei's review? Or is this ready to merge?

This PR is standalone. It just adds a model to the list of models we currently support. We can wait. The most compelling PR for me is PR#105 because it always causes conflicts when I update it from the main branch

hydragnn/models/SAGEStack.py
```diff
@@ -0,0 +1,59 @@
...
+            dropout,
+            num_conv_layers,
+            num_nodes,
+        )
+
+    def get_conv(self, input_dim, output_dim):
+        return SAGEConv(
+            in_channels=input_dim,
+            out_channels=output_dim,
+            root_weight=False,
```
pzhanggit, line 55: Any reason to  set `root_weight=False` and overwrite the default model formula?

allaffa: Yes. I have not been completely honest with you guys. There are other hidden branches behind this PR that seems standalone XD
I am deliberately avoiding any **non-linearity** whenever I can to try to build a neural network that predicts properties that scale linearly with the size of the lattice.
I can introduce an additional input parameter to make this optional.

pzhanggit: I hear you, but if I understand correctly, the`root_weight` here is not about square root of any weights. It is about the first part of 
image

allaffa: You are right. I misinterpreted the description. I just looked at the implementation 
[https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/nn/conv/sage_conv.html#SAGEConv](url)
Screen Shot 2022-03-08 at 11 14 09 PM

tests/test_graphs.py
```diff
@@ -122,6 +122,7 @@ def unittest_train_model(model_type, ci_input, use_lengths, overwrite_data=False
 
     # Set RMSE and sample MAE/max error thresholds
     thresholds = {
+        "SAGE": [0.45, 0.45, 0.90],
```
pzhanggit, line 125: It seems SAGE is doing okay in the unit test after the changes. Maybe we can use the same error threshold as the other models?

pzhanggit (approved):

allaffa:
Now I am afraid of merging anything before PR#105. XD

PR #110, opened 2022-03-07
Add initial jenkins CI

streeve:
Not yet enabled on the Jenkins server

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -1,0 +1,27 @@
+pipeline {
+    triggers {
+        issueCommentTrigger('.*test this please.*')
+    }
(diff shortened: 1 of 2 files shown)


allaffa:
Where can I run the code to validate that it works?

streeve:
> Where can I run the code to validate that it works?

A new check will show up when this is connected - I forgot to add the PR description

allaffa:
> > Where can I run the code to validate that it works?
> 
> A new check will show up when this is connected - I forgot to add the PR description

Should we convert the PR to draft then?

PR #111, opened 2022-03-07
Add ability to freeze convolution layers

streeve:
This is potentially useful for PI3NN runs. Currently made this an optional input and did not explicitly add to any existing config files


streeve:
This fails because for GAT we need to know whether to average or concat - can I just concat?

streeve:
I assumed yes

hydragnn/models/GATStack.py
```diff
@@ -82,7 +82,7 @@ def _init_node_conv(self):
         self.hidden_dim_node = self.config_heads["node"]["dim_headlayers"]
         # In this part, each head has same number of convolutional layers, but can have different output dimension
         self.convs_node_hidden.append(
-            self.get_conv(self.hidden_dim, self.hidden_dim_node[0], True)
+            self.get_conv(self.hidden_dim, self.hidden_dim_node[0])
```
allaffa, line 85: See comment above.

hydragnn/models/GATStack.py
```diff
@@ -106,7 +105,7 @@ def _init_node_conv(self):
             )
             self.batch_norms_node_output.append(BatchNorm(self.head_dims[ihead]))
 
-    def get_conv(self, input_dim, output_dim, concat):
+    def get_conv(self, input_dim, output_dim, concat=True):
```
allaffa, line 108: See comment above.

hydragnn/models/GATStack.py
```diff
@@ -56,10 +56,10 @@ def __init__(
     def _init_conv(self):
         """Here this function overwrites _init_conv() in Base since it has different implementation
         in terms of dimensions due to the multi-head attention"""
-        self.convs.append(self.get_conv(self.input_dim, self.hidden_dim, True))
+        self.convs.append(self.get_conv(self.input_dim, self.hidden_dim))
```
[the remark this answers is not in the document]
allaffa: No need to split. Thanks for the explanation

allaffa (approved):

hydragnn/models/Base.py
```diff
@@ -81,7 +81,12 @@ def __init__(
         ):
             self.use_edge_attr = True
 
+        # Option to only train final property layers.
+        self.freeze_conv = True
```
streeve, line 85: Oops

pzhanggit:
> > I assumed yes
> 
> Right now, the models developed by @pzhanggit always use concat. Pei can you confirm?

@allaffa  yes, concat has been used. The changes @streeve made optimize the implementation and also enable a consistent function form with other models if I understand correctly

hydragnn/models/GATStack.py
```diff
@@ -106,7 +105,7 @@ def _init_node_conv(self):
             )
```
streeve, line 105: To be clear, we do not concat on the line above this

hydragnn/models/Base.py
streeve: Fixed

hydragnn/models/Base.py
```diff
@@ -93,6 +98,11 @@ def _init_conv(self):
             self.convs.append(conv)
             self.batch_norms.append(BatchNorm(self.hidden_dim))
 
+    def _freeze_conv(self):
+        conv = self.get_conv(self.input_dim, self.hidden_dim)
```
pzhanggit, line 102: Does not look right to me. At this stage, stacked GNN layers are already defined in `self.convs`. To freeze them, should not you go inside `self.convs` and freeze the parameters there?
What currently done is define a new GNN layer (which is not used by any model) and free its parameters.
Let me know if I completely misunderstand it. Thanks.

hydragnn/models/Base.py
```diff
@@ -99,8 +99,7 @@ def _init_conv(self):
             self.batch_norms.append(BatchNorm(self.hidden_dim))
 
     def _freeze_conv(self):
-        conv = self.get_conv(self.input_dim, self.hidden_dim)
-        for param in conv.parameters():
+        for param in self.convs:
```
pzhanggit, line 102: @streeve  Why the two versions of getting parameters are different? Do they both work? I was expecting something like
`for conv in self.convs:
        for param in conv.paramters():
            param.requires_grad = False` Could you verify? Thanks.

hydragnn/models/Base.py
```diff
@@ -81,7 +81,12 @@ def __init__(
         ):
             self.use_edge_attr = True
 
+        # Option to only train final property layers.
+        self.freeze_conv = False
+
         self._init_conv()
+        if self.freeze_conv:
```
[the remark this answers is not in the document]
streeve: Done

hydragnn/models/Base.py
streeve: Fixed, thanks

hydragnn/models/GATStack.py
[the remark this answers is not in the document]
streeve: FYI this was removed since it wasn't necessary when accessing the member variables directly

hydragnn/models/Base.py
```diff
@@ -81,7 +81,12 @@ def __init__(
         ):
             self.use_edge_attr = True
 
+        # Option to only train final property layers.
+        self.freeze_conv = False
```
pzhanggit, line 85: getting `freeze_conv` from `create_model` is missing?

streeve: fixed, thanks for finding another mistake

streeve:
Ready for final review

pzhanggit (approved):
Look good to me, thanks!

PR #113, opened 2022-03-08
opening of raw data file moved into __transform method

allaffa:
This will facilitate the conclusion of additional `__transform methods` to support multiple data formats


allaffa:
@streeve @pzhanggit 
I think this can be quickly reviewed :-)

pzhanggit (approved):

PR #115, opened 2022-03-08
delayed transfer of data from host to device

allaffa:
Delaying transfer of data from host to device is needed when the dataset is too large to fit entirely onto a single GPU.

### DIFF
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -15,6 +15,7 @@ import torch.nn.functional as F
 from torch_geometric.nn import global_mean_pool, BatchNorm
 from torch.nn import GaussianNLLLoss
 import sys
+from hydragnn.utils.distributed import get_device
 
 
 class Base(Module):
@@ -35,6 +36,7 @@ class Base(Module):
         num_nodes: int = None,
     ):
         super().__init__()
+        self.device = get_device()
         self.input_dim = input_dim
         self.hidden_dim = hidden_dim
         self.dropout = dropout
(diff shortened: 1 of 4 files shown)


jychoi-hpc:
I will check:
* different places of moving data from CPU to GPU.
* Mak a (config) option for users to specify explicitly (for performance reason)

jychoi-hpc:
I tested with `examples/lsms` case and found this PR reduces the loading time from 45 sec to 30 sec on Summit. Per-batch data moving from CPU to GPU (line 204 in `Base.py`) was not a big deal.

My observation is the function `update_predicted_values`, which is called after serial data loading, works better on CPU than GPU and thus It is better not call `data.to(device)` as in this PR.

I think this PR can help the GPU out-of-memory error in some degree. Ideally, if we can avoid any this type of post processing calculations after data loading from disk would be the best scenario. We can discuss further after this merge.

jychoi-hpc (approved):

streeve (changes requested):
Review didn't show up for some reason

jychoi-hpc:
I found this PR is necessary for the distributed processing with Ising and OGB. So, I put the most of changes in this PR to #135  and tested working there.  Can we drop this PR and, instead, work on #135?

allaffa:
> I found this PR is necessary for the distributed processing with Ising and OGB. So, I put the most of changes in this PR to #135 and tested working there. Can we drop this PR and, instead, work on #135?

@jychoi-hpc 
Yes. I am going to close it.

PR #119, opened 2022-03-21
Update on checkpointing

jychoi-hpc:
This is an update for checkpointing and restarting. 

Now it saves both model and optimizer states and loads when restarting. It works in a distributed setting too.


streeve (approved):
Looks great!

pzhanggit (approved):

allaffa (approved):

PR #124, opened 2022-03-24
Scatter plot for vector outputs

pzhanggit:


hydragnn/postprocess/visualizer.py
```diff
@@ -636,12 +687,13 @@ def create_scatter_plots(
     ):
         """Creates scatter plots for all head predictions."""
         for ihead in range(self.num_heads):
-            if self.head_dims[ihead] == 3:
+            if self.head_dims[ihead] > 1:
                 # vector output
-                self.create_scatter_plot_nodes_vec(
```
streeve, line 641: What happens for vector node properties?

pzhanggit: It is still there. We do not use it for now. It was designed specifically for node feature vector with three components like magnetic moment in alloy datasets.

streeve: I misunderstood this last week. Planning to push naming changes and then I think this is ready to merge

streeve (approved):
Ready to merge if you are happy with the new function names

PR #127, opened 2022-03-30
bug fixed on calculation of training loss function and new loss functions introduced

allaffa:


pzhanggit (changes requested):
Some general comments:
Not sure if it is a bug, since as the name `loss_rmse` suggests, it was rmse loss. Maybe instead of removing it completely, we keep it as an option, in addition to the MSE. This can also help us reproduce the old results in existing applications.

pzhanggit:
> I would call it a bug because the RMSE was calculated wrong. We can keep a function to calculate final RMSE, but this should not match previous results

Two things here: RMSE loss based on mini-batch VS. RMSE of entire dataset. It will only give you the former. So if you use it for the later then sure it is a bug.

allaffa:
@pzhanggit @streeve 
It's ready for review

pzhanggit (changes requested):
Looks alright to me. A few general comments:
1. Could you make `loss_function_type` optional to users and set default to be 'mse'? Maybe add a few lines in update config function. In this way, our old cases will not break.
2. Add "rmse" as one of the `loss_function_selection`, for two reasons. We can reproduce our old results. And also after our discussion, I think we agree it is a practically legitimate loss function (?).
3. Change variable names like `val_mse` in train_validate_test.py to be someting general, since it is not necessarily mse now.

hydragnn/models/Base.py
```diff
@@ -55,6 +57,7 @@ def __init__(
         self.convs_node_output = ModuleList()
         self.batch_norms_node_output = ModuleList()
 
+        self.loss_function = loss_function_selection(loss_function_type)
```
streeve, line 60: Very nice

streeve (commented):
I would approve with Pei's suggested changes. She makes very good points

tests/test_model_loadpred.py
```diff
@@ -63,14 +64,16 @@ def unittest_model_prediction(config):
         )
 
 
-# test loading and predictiong of a saved model from previous training
-def pytest_model_loadpred():
+# test loading and prediction of a saved model from previous training
+@pytest.mark.parametrize("loss_function_type", ["mae", "mse", "rmse"])
```
pzhanggit, line 68: Separate the loss function test in a different test? This test is intended for loading and prediction of a pretrained model.

pzhanggit (commented):
Thanks. It looks good to me. One comment is about the unit test of loss function. It is better to put it in a separate test IMO.

tests/test_model_loadpred.py
allaffa: I do not want to make the test heavy weight. You are calling the loss function in this test. I would like just to test if the loss function crashes or not. I thought this would be a good place. But I can change it anyway

pzhanggit: Thanks. The other reason to separate them is that the loss function` ["mae", "mse", "rmse"]` is not going to be checked as desired. The code will check first if a pre-trained model and dataset exist in `log_name`, and if they do, it will load the the old case with whatever loss function used there. If really want to check the loss function here, we probably can add `loss_function_type` as part of `log_name`.

allaffa: @pzhanggit I created a separate `test_loss_function.py`
I only run 2 epochs, definitely not sufficient to check the accuracy, but enough to see if the loss function crashes or not.

allaffa: @streeve 
If I want to check only if a capability crashes or not, without necessarily bothering about the accuracy, is it ok to create a python test without any specific assert?

streeve: Yes, that can be reasonable. That's all the example "tests" do right now

allaffa: Good. So the the new test should be setup correctly.

streeve:
@allaffa I pushed my suggested changes here. I will approve if it passes

allaffa:
> @allaffa I pushed my suggested changes here. I will approve if it passes

@streeve Thanks. It passed. Go ahead and approve.

streeve (approved):

pzhanggit (approved):

PR #129, opened 2022-03-31
Add a test to load trained model and run prediction for a single sample

pzhanggit:


hydragnn/models/Base.py
```diff
@@ -244,7 +244,10 @@ def forward(self, data):
 
         #### multi-head decoder part####
         # shared dense layers for graph level output
-        x_graph = global_mean_pool(x, batch.to(x.device))
+        if batch is None:
```
allaffa, line 247: Nice! I remember having similar issues for other codes every time I called **global_mean_pool** on a single data sample.

tests/test_graph_pred.py
```diff
@@ -0,0 +1,53 @@
...
+    log_name = hydragnn.utils.config_utils.get_log_name_config(config)
+    hydragnn.utils.model.load_existing_model(model, log_name)
+
+    isample = random.randrange(len(test_loader.dataset))
+    graph_sample = test_loader.dataset[isample]
+    true_values = graph_sample.y.squeeze()
+    predicted_values = model(graph_sample)
+    yloc = graph_sample.y_loc.squeeze()
+
+    thresholds = 0.75
```
allaffa, line 45: This threshold every high because the goal of this test is NOT to test the accuracy, but just to make sure that a model is correctly loaded and deployed for predictions. Correct?

allaffa (commented):
Just questions to clarify the set up of the unit test.

tests/test_graph_pred.py
pzhanggit: We want to check model loading and deployment as well as accuracy. The threshold is set to be equal to the maximum sample error we had in https://github.com/ORNL/HydraGNN/blob/215b5e8c93ff945e7da8951adc7fe9f5e3676c9c/tests/test_graphs.py#L126, since we are checking a single randomly selected sample. We could add one more to check the entire test set.

tests/test_graph_pred.py
```diff
@@ -0,0 +1,53 @@
...
+##############################################################################
+import os, json
+import torch
+import random
+import hydragnn
+
+
+def unittest_model_pred(model_dir):
+    _, _ = hydragnn.utils.get_comm_size_and_rank()
+    # Read in config settings and override model type.
```
streeve, line 19: Not overriding?

tests/test_graph_pred.py
```diff
@@ -0,0 +1,53 @@
...
+#                                                                            #
+# SPDX-License-Identifier: BSD-3-Clause                                      #
+##############################################################################
+import os, json
+import torch
+import random
+import hydragnn
+
+
+def unittest_model_pred(model_dir):
```
streeve, line 17: `unittest_model_prediction` + comment? This is just for testing prediction without training?

tests/test_graph_pred.py
```diff
@@ -0,0 +1,53 @@
...
+
+    (
+        train_loader,
+        val_loader,
+        test_loader,
+        sampler_list,
+    ) = hydragnn.preprocess.load_data.dataset_loading_and_splitting(config=config)
+
+    model = hydragnn.models.create.create_model_config(
+        config=config["NeuralNetwork"]["Architecture"],
```
streeve, line 32: This will have to change based on #127

tests/test_graphs.py
```diff
@@ -16,6 +16,7 @@
 import shutil
 
 import hydragnn, tests
+from .test_graph_pred import unittest_model_pred
```
streeve, line 19: Let's just do all of this in your new file

tests/test_graph_pred.py
```diff
@@ -0,0 +1,53 @@
...
+    config_file = os.path.join(model_dir, "config.json")
+    with open(config_file, "r") as f:
+        config = json.load(f)
+
+    (
+        train_loader,
+        val_loader,
+        test_loader,
+        sampler_list,
+    ) = hydragnn.preprocess.load_data.dataset_loading_and_splitting(config=config)
```
streeve, line 29: Does this rely on another test having created the data? Or am I missing something

streeve (changes requested):

tests/test_graph_pred.py
pzhanggit: Right.

tests/test_graph_pred.py
pzhanggit: Why? I do not see any changes needed after merging https://github.com/ORNL/HydraGNN/pull/127.

tests/test_graph_pred.py
pzhanggit: Yes, this test is for model loading and prediction check only.

tests/test_graphs.py
pzhanggit: Sure. One thing is that we want this test to run after the model training done in https://github.com/ORNL/HydraGNN/blob/bf1915c9ac3db508067f365289b674cf201bb7f5/tests/test_graphs.py#L183. Is a place where I can specify the order of unit tests?

tests/test_graph_pred.py
pzhanggit: Could you explain more?

tests/test_graphs.py
streeve: Let's use https://pytest-dependency.readthedocs.io/en/latest/usage.html#basic-usage :)

tests/test_graph_pred.py
pzhanggit: Removed.

tests/test_graphs.py
pzhanggit: It did not work as we expected, added a manual checking.

streeve (approved):
Looks good. Maybe we can revisit the pytest-dependency later on

PR #130, opened 2022-04-04
option to select between different optimizers added

allaffa:


hydragnn/utils/optimizer.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def selected_optimizer(model, config):
+    optimizer = None
+
+    if config["optimizer"] == "SGD":
+        optimizer = torch.optim.SGD(model.parameters(), lr=config["learning_rate"])
+    elif config["optimizer"] == "Adam":
+        optimizer = torch.optim.Adam(model.parameters(), lr=config["learning_rate"])
+    elif config["optimizer"] == "Adadelta":
```
streeve, line 11: typo

hydragnn/utils/optimizer.py
```diff
@@ -0,0 +1,32 @@
...
+        optimizer = torch.optim.Adamdelta(
+            model.parameters(), lr=config["learning_rate"]
+        )
+    elif config["optimizer"] == "Adagrad":
```
streeve, line 15: typo

streeve (changes requested):
Like Pei mentioned before, adding AdamW as default to keep backwards compatibility will be good (in `update_config` I think?)

hydragnn/utils/optimizer.py
allaffa: Thanks.

hydragnn/utils/optimizer.py
allaffa: Thanks.

allaffa:
> Like Pei mentioned before, adding AdamW as default to keep backwards compatibility will be good (in `update_config` I think?)

`run_training.py` calls `update_config`
If I pass "AdamW" into `update_config`, this would imply that the optimizer chosen gets overwritten every time `run_training.py` is called. Correct?

allaffa:
> Like Pei mentioned before, adding AdamW as default to keep backwards compatibility will be good (in `update_config` I think?)
@streeve 
Valid point. Done.

hydragnn/utils/optimizer.py
```diff
@@ -0,0 +1,30 @@
+import torch
+
+
+def selected_optimizer(model, config):
```
streeve, line 4: How about `select_optimizer` or `choose_optimizer`

allaffa: I changed the name to `select_optimizer`

allaffa (commented):
I changed the name of the utility as suggested by @streeve

streeve (approved):

PR #132, opened 2022-04-11
Add  LAMB optimizer

allaffa:


allaffa:
@streeve 
I think I finished addressing your requests. 
Take a look at `test_optimizer.py` and tell me if you can think of a more convenient way to make the execution of the LAMB optimizer optional to avoid the tests to fail in absence of GPUs.

tests/test_optimizer.py
```diff
@@ -0,0 +1,120 @@
...
+                if not os.listdir(data_path):
+                    tests.deterministic_graph_data(
+                        data_path, number_configurations=num_samples
+                    )
+
+    config["NeuralNetwork"]["Training"]["optimizer"] = optimizer_type
+    config["NeuralNetwork"]["Training"]["num_epoch"] = 2
+
+    # Make sure training works with each optimizer.
+    if config["NeuralNetwork"]["Training"]["optimizer"] == "LAMB":
```
streeve, line 95: I would just leave lamb out of the tests for now (maybe a comment saying it should be tested on GPU). Otherwise, there are options to expect a test to fail

hydragnn/utils/optimizer.py
```diff
@@ -16,14 +23,16 @@ def select_optimizer(model, config):
         optimizer = torch.optim.Adamax(model.parameters(), lr=config["learning_rate"])
     elif config["optimizer"] == "AdamW":
         optimizer = torch.optim.AdamW(model.parameters(), lr=config["learning_rate"])
-    elif config["optimizer"] == "LBFGS":
-        optimizer = torch.optim.LBFGS(model.parameters(), lr=config["learning_rate"])
-    elif config["optimizer"] == "SparseAdam":
-        optimizer = torch.optim.SparseAdam(
+    elif config["optimizer"] == "RMSprop":
+        optimizer = torch.optim.RMSprop(model.parameters(), lr=config["learning_rate"])
+    elif config["optimizer"] == "LAMB":
```
streeve, line 28: Should we use FusedLamb to match the torch option?

allaffa: Very good point

PR #135, opened 2022-04-27
Distributed data processing

jychoi-hpc:
This is the update I made to run Ising model using multi-GPUs on Summit and Perlmutter.

Major updates:
* Add distributed raw data/serial data loading (each process will read different set of data)
* Parallel normalization and min-max (using DDP to aggregate)
* Using Adios2 for Ising

It looks like many changes but I think this is necessary for distributed processing. Please feel free to give any comments.

### DIFF
--- a/examples/ising_model/create_configurations.py
+++ b/examples/ising_model/create_configurations.py
@@ -16,7 +16,7 @@ def write_to_file(total_energy, atomic_features, count_config, dir):
     for index in range(0, atomic_features.shape[0]):
         numpy_row = atomic_features[index, :]
         numpy_string_row = np.array2string(
-            numpy_row, precision=2, separator="\t", suppress_small=True
+            numpy_row, separator="\t", suppress_small=True
         )
         filetxt += "\n" + numpy_string_row.lstrip("[").rstrip("]")
 
@@ -43,9 +43,9 @@ def E_dimensionless(config, L, spin_function, scale_spin):
     number_nodes = L ** 3
     positions = np.zeros((number_nodes, 3))
     atomic_features = np.zeros((number_nodes, 5))
-    for z in range(L):
+    for x in range(L):
         for y in range(L):
-            for x in range(L):
+            for z in range(L):
                 positions[count_pos, 0] = x
                 positions[count_pos, 1] = y
                 positions[count_pos, 2] = z
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -228,7 +228,7 @@ class Base(Module):
             self.heads_NN.append(head_NN)
 
     def forward(self, data):
-        data.to(self.device)
+        data = data.to(self.device)
         x, edge_index, batch = (
             data.x,
             data.edge_index,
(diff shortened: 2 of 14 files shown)


streeve (commented):
Like we discussed at the meeting we can switch to DDP required an simplify some of these changes (could be separate if you prefer)

hydragnn/models/Base.py
```diff
@@ -226,6 +228,7 @@ def _multihead(self):
             self.heads_NN.append(head_NN)
 
     def forward(self, data):
+        data = data.to(self.device)
```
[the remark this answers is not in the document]
jychoi-hpc: I found previously we move data from cpu to gpu in serialized data loader: https://github.com/ORNL/HydraGNN/blob/b48f75cd3fe1b0d03bae9af3e6bdc2bb29f8b9c6/hydragnn/preprocess/serialized_dataset_loader.py#L140
I think moving data from cpu to gpu here makes more sense to me. It also works with the case when serialized loader is not used (like Adios).

examples/ising_model/create_configurations.py
```diff
@@ -43,9 +43,9 @@ def E_dimensionless(config, L, spin_function, scale_spin):
     number_nodes = L ** 3
     positions = np.zeros((number_nodes, 3))
     atomic_features = np.zeros((number_nodes, 5))
-    for z in range(L):
+    for x in range(L):
```
allaffa, line 46: Can you explain why you had to invert the order of these nested loop?

hydragnn/models/Base.py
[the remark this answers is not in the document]
allaffa: Previously, @streeve put this line inside the serialized data loader because small data can fit entirely on the GPU. 
I am on to lose some accuracy on small data by benefitting on large data that needs HPC resources. Because there is where we really care about performance. 
@pzhanggit what do you think?

hydragnn/preprocess/load_data.py
```diff
@@ -50,15 +52,28 @@ def dataset_loading_and_splitting(config: {}):
     )
 
 
+def worker_init_fn(worker_id):
```
allaffa, line 55: Did you add this function with the idea that the functionalities of this function could be expanded in the future? 
It seems like the only thing it does is writing a line on the log file. Is it true?

hydragnn/preprocess/utils.py
```diff
@@ -16,16 +16,54 @@
 import ase
 import ase.neighborlist
 
-
+## This function can be slow if dataset is too large. Use with caution.
```
allaffa, line 19: Is this where @streeve was mentioning the idea of avoiding multiple replicas of the same function?

hydragnn/utils/model.py
```diff
@@ -90,14 +81,42 @@ def load_existing_model(model, model_name, path="./logs/", optimizer=None):
         optimizer.load_state_dict(checkpoint["optimizer_state_dict"])
 
 
+## This function may cause OOM if dataset is too large
+## to fit in a single GPU (i.e., with DDP). Use with caution.
+## Recommend to use calculate_PNA_degree_dist
 def calculate_PNA_degree(dataset: [Data], max_neighbours):
-    deg = torch.zeros(max_neighbours + 1, dtype=torch.long).to(get_device())
+    deg = torch.zeros(max_neighbours + 1, dtype=torch.long)
```
allaffa, line 88: I assume this is another point where @streeve was mentioning the idea of avoiding function replicas. Correct?

jychoi-hpc:
Let's drop this PR for now. Will make a PR with the current OGB/CSCE update.

PR #136, opened 2022-05-02
DDP bug fix

pzhanggit:
Major features/changes:
1. Fixed a bug in DDP model. The DDP wrapped model was unwrapped (which is not supposed to be) before training. As a result, every process trains its own model with a subset of dataset and there is no synchronization between them. 
2. Fixed a bug about map_location in model loading
3. Added reduce/gather feature for loss log and results collection
4. Added SyncBatchNorm 
MAE_GPUs_comp_test_ddpbug

Checked that the gradients and model parameters on rank 0 and rank 1 are the same
gradient_check
parameters_check

### DIFF
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -198,6 +198,7 @@ class Base(Module):
                 self.node_NN_type = self.config_heads["node"]["type"]
                 head_NN = ModuleList()
                 if self.node_NN_type == "mlp" or self.node_NN_type == "mlp_per_node":
+                    self.num_mlp = 1 if self.node_NN_type == "mlp" else self.num_nodes
                     assert (
                         self.num_nodes is not None
                     ), "num_nodes must be positive integer for MLP"
@@ -205,7 +206,7 @@ class Base(Module):
                     head_NN = MLPNode(
                         self.hidden_dim,
                         self.head_dims[ihead],
-                        self.num_nodes,
+                        self.num_mlp,
                         self.hidden_dim_node,
                         self.config_heads["node"]["type"],
                     )
@@ -329,15 +330,15 @@ class Base(Module):
 
 
 class MLPNode(Module):
-    def __init__(self, input_dim, output_dim, num_nodes, hidden_dim_node, node_type):
+    def __init__(self, input_dim, output_dim, num_mlp, hidden_dim_node, node_type):
         super().__init__()
         self.input_dim = input_dim
         self.output_dim = output_dim
-        self.num_nodes = num_nodes
         self.node_type = node_type
+        self.num_mlp = num_mlp
 
         self.mlp = ModuleList()
-        for inode in range(self.num_nodes):
+        for _ in range(self.num_mlp):
             denselayers = []
             denselayers.append(Linear(self.input_dim, hidden_dim_node[0]))
             denselayers.append(ReLU())
@@ -348,8 +349,6 @@ class MLPNode(Module):
                 denselayers.append(ReLU())
             denselayers.append(Linear(hidden_dim_node[-1], output_dim))
             self.mlp.append(Sequential(*denselayers))
-            if self.node_type == "mlp":
-                break
 
     def node_features_reshape(self, x, batch):
         """reshape x from [batch_size*num_nodes, num_features] to [batch_size, num_features, num_nodes]"""
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -94,7 +94,6 @@ class Base(Module):
         self._init_conv()
         if self.freeze_conv:
             self._freeze_conv()
-        self._init_node_conv()
         self._multihead()
         if self.initial_bias is not None:
             self._set_bias()
@@ -169,6 +168,11 @@ class Base(Module):
                 denselayers.append(ReLU())
             self.graph_shared = Sequential(*denselayers)
 
+        if "node" in self.config_heads:
+            self.num_conv_layers_node = self.config_heads["node"]["num_headlayers"]
+            self.hidden_dim_node = self.config_heads["node"]["dim_headlayers"]
+            self._init_node_conv()
+
         inode_feature = 0
         for ihead in range(self.num_heads):
             # mlp for each head output
@@ -193,8 +197,6 @@ class Base(Module):
             elif self.head_type[ihead] == "node":
                 self.node_NN_type = self.config_heads["node"]["type"]
                 head_NN = ModuleList()
-                self.num_conv_layers_node = self.config_heads["node"]["num_headlayers"]
-                self.hidden_dim_node = self.config_heads["node"]["dim_headlayers"]
                 if self.node_NN_type == "mlp" or self.node_NN_type == "mlp_per_node":
                     assert (
                         self.num_nodes is not None
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -125,13 +125,16 @@ class Base(Module):
         # two ways to implement node features from here:
         # 1. one graph for all node features
         # 2. one graph for one node features (currently implemented)
+        if (
+            "node" not in self.config_heads
+            or self.config_heads["node"]["type"] != "conv"
+        ):
+            return
         node_feature_ind = [
             i for i, head_type in enumerate(self.head_type) if head_type == "node"
         ]
         if len(node_feature_ind) == 0:
             return
-        self.num_conv_layers_node = self.config_heads["node"]["num_headlayers"]
-        self.hidden_dim_node = self.config_heads["node"]["dim_headlayers"]
         # In this part, each head has same number of convolutional layers, but can have different output dimension
         self.convs_node_hidden.append(
             self.get_conv(self.hidden_dim, self.hidden_dim_node[0])
@@ -190,6 +193,8 @@ class Base(Module):
             elif self.head_type[ihead] == "node":
                 self.node_NN_type = self.config_heads["node"]["type"]
                 head_NN = ModuleList()
+                self.num_conv_layers_node = self.config_heads["node"]["num_headlayers"]
+                self.hidden_dim_node = self.config_heads["node"]["dim_headlayers"]
                 if self.node_NN_type == "mlp" or self.node_NN_type == "mlp_per_node":
                     assert (
                         self.num_nodes is not None
@@ -341,6 +346,8 @@ class MLPNode(Module):
                 denselayers.append(ReLU())
             denselayers.append(Linear(hidden_dim_node[-1], output_dim))
             self.mlp.append(Sequential(*denselayers))
+            if self.node_type == "mlp":
+                break
 
     def node_features_reshape(self, x, batch):
         """reshape x from [batch_size*num_nodes, num_features] to [batch_size, num_features, num_nodes]"""
--- a/hydragnn/postprocess/visualizer.py
+++ b/hydragnn/postprocess/visualizer.py
@@ -116,9 +116,10 @@ class Visualizer:
         y_label=None,
         xylim_equal=False,
     ):
+
         ax.scatter(x, y, s=s, edgecolor="b", marker=marker, facecolor="none")
 
-        ax.set_title(title)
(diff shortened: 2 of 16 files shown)


allaffa:
@pzhanggit 
I am running the code on FePt. 
I will let you know what I obtain as a result. 
While I am at it, do you want me to run the PR on other datasets that you did not run yourself?

pzhanggit:
> @pzhanggit I am running the code on FePt. I will let you know what I obtain as a result. While I am at it, do you want me to run the PR on other datasets that you did not run yourself?

Thanks, @allaffa. What datasets you have in mind? So far all I ran is the dftb version of qm9 and in the single head mode. Could you run one case with multihead?

jychoi-hpc:
This is good. I will run a test on Summit. Which is the part doing "unwrapping" ddp model which is not supposed to happen?

hydragnn/train/train_validate_test.py
```diff
@@ -46,16 +48,16 @@ def train_validate_test(
...
-    total_loss_test = [None] * num_epoch
+    total_loss_train = torch.zeros(num_epoch, device=device)
+    total_loss_val = torch.zeros(num_epoch, device=device)
+    total_loss_test = torch.zeros(num_epoch, device=device)
     # loss tracking for each head/task
-    task_loss_train = [None] * num_epoch
-    task_loss_test = [None] * num_epoch
-    task_loss_val = [None] * num_epoch
-
-    model = get_model_or_module(model)
```
pzhanggit, line 58: @jychoi-hpc  the unwrapping here

allaffa:
@pzhanggit 
Now we print the full pytorch tensor on the screen, which causes less digits to be printed out for the individual loss functions on each task. 
We should roll back to there same print formatting as before, maybe calling `torch.item()` instead of printing the entire tensor

hydragnn/utils/model.py
```diff
@@ -75,7 +76,7 @@ def load_existing_model(model, model_name, path="./logs/", optimizer=None):
     """Load both model and optimizer state from a single checkpoint file"""
     _, world_rank = get_comm_size_and_rank()
     path_name = os.path.join(path, model_name, model_name + ".pk")
-    map_location = {"cuda:%d" % 0: "cuda:%d" % world_rank}
+    map_location = {"cuda:%d" % 0: get_device_name()}
```
streeve, line 79: Maybe we rename these `get_local_device_name` and `get_global_size_and_rank`?

hydragnn/postprocess/visualizer.py
```diff
@@ -282,6 +283,8 @@ def create_parity_plot_and_error_histogram_scalar(
     ):
         """Creates scatter plots for true_values and predicted values of variable varname at iepoch."""
 
+        true_values = true_values.cpu().detach().numpy()
```
streeve, line 286: Good idea

hydragnn/train/train_validate_test.py
```diff
@@ -102,7 +105,11 @@ def train_validate_test(
             )
         val_loss, val_taskserr = validate(val_loader, model, verbosity)
         test_loss, test_taskserr, true_values, predicted_values = test(
-            test_loader, model, verbosity
+            test_loader,
+            model,
+            verbosity,
+            reduce_gpus=True,
```
streeve, line 111: reduce_ddp? or reduce_ranks?

hydragnn/train/train_validate_test.py
```diff
@@ -205,6 +218,67 @@ def get_head_indices(model, data):
...
+    if dist.get_world_size() > 1:
+        dist.all_reduce(total_error, op=dist.ReduceOp.SUM)
+        total_error = total_error / dist.get_world_size()
+        dist.all_reduce(tasks_error, op=dist.ReduceOp.SUM)
+        tasks_error = tasks_error / dist.get_world_size()
+    return total_error, tasks_error
+
+
+@torch.no_grad()
+def gather_head_values(head_values):
```
streeve, line 243: gather_array_ddp?

hydragnn/train/train_validate_test.py
```diff
@@ -205,6 +218,67 @@ def get_head_indices(model, data):
     return head_index
 
 
+@torch.no_grad()
+def reduce_loss_stat(total_loss_list, task_loss_list):
```
streeve, line 222: Can we reuse one `reduce_values_ddp` function that takes a single list and returns the reduction or updates in place? Unless we really need both options

hydragnn/train/train_validate_test.py
```diff
@@ -205,6 +218,67 @@ def get_head_indices(model, data):
...
+    # gather error statistics across all processes
+    if dist.get_world_size() > 1:
+        for total_error, tasks_error in zip(total_loss_list, task_loss_list):
+            dist.all_reduce(total_error, op=dist.ReduceOp.SUM)
```
streeve, line 226: We discussed this a bit already, but I don't see a reason to all_gather instead of gather to rank zero (here and below)

hydragnn/utils/distributed.py
```diff
@@ -213,7 +213,7 @@ def is_model_distributed(model):
     return isinstance(model, torch.nn.parallel.distributed.DistributedDataParallel)
 
 
-def get_distributed_model(model, verbosity=0):
+def get_distributed_model(model, verbosity=0, SyncBatchNorm=False):
```
streeve, line 216: sync_batch_norm

hydragnn/models/Base.py
```diff
@@ -94,7 +94,6 @@ def __init__(
         self._init_conv()
         if self.freeze_conv:
             self._freeze_conv()
-        self._init_node_conv()
```
streeve, line 97: This seems unrelated

hydragnn/train/train_validate_test.py
streeve: We should remove that function entirely, right?

streeve (commented):
This looks like a lot of work and I'm very glad the results make more sense once we're actually using DDP :)

allaffa (approved):

hydragnn/postprocess/visualizer.py
```diff
@@ -116,9 +116,10 @@ def __scatter_impl(
         y_label=None,
         xylim_equal=False,
     ):
+
         ax.scatter(x, y, s=s, edgecolor="b", marker=marker, facecolor="none")
 
-        ax.set_title(title)
+        ax.set_title(title + ", len=" + str(len(x)))
```
allaffa, line 122:
```suggestion
        ax.set_title(title + ", number data samples = " + str(len(x)))
```

hydragnn/utils/distributed.py
```diff
@@ -213,12 +213,27 @@ def is_model_distributed(model):
     return isinstance(model, torch.nn.parallel.distributed.DistributedDataParallel)
 
 
-def get_distributed_model(model, verbosity=0):
-    device_name = get_device_name(verbosity)
+def get_distributed_model(model, verbosity=0, SyncBatchNorm=False):
+    device_name = get_device_name(verbosity_level=verbosity)
+
+    world_size, world_rank = get_comm_size_and_rank()
+    print(
```
allaffa, line 220: I would use print verbosity here. What do you think?

hydragnn/train/train_validate_test.py
```diff
@@ -216,89 +290,97 @@ def train(
...
-        total_error += loss.item() * data.num_graphs
-        num_samples_local += data.num_graphs
-        for itask in range(len(tasks_loss)):
-            tasks_error[itask] += tasks_loss[itask].item() * data.num_graphs
-    return (
-        total_error / num_samples_local,
-        tasks_error / num_samples_local,
-    )
+        with torch.no_grad():
+            total_error += loss * data.num_graphs
```
allaffa, line 310: any reason why you do not access the .item() anymore? 
The older printout to the screen looks better. More digits are printed on the screen with the old way

allaffa (changes requested):

hydragnn/train/train_validate_test.py
pzhanggit: Agreed. We will open an issue and make another PR as we discussed

hydragnn/models/Base.py
pzhanggit: It is relevant. Previously, conv for node is defined even it is not used in the model. However, DDP does not support unused parameters. So we change it to only define it when used.

hydragnn/utils/distributed.py
pzhanggit: Probably no difference here since we want to print on every rank

hydragnn/train/train_validate_test.py
pzhanggit: Changed

hydragnn/train/train_validate_test.py
pzhanggit: Not sure. It is used also in the model saving. Probably related to the non-DDP mode we were discussing earlier. Let's keep it for now.

hydragnn/utils/model.py
pzhanggit: Not relevant to this PR?

hydragnn/train/train_validate_test.py
pzhanggit: Done

hydragnn/train/train_validate_test.py
pzhanggit: Done

hydragnn/train/train_validate_test.py
pzhanggit: Done

hydragnn/utils/distributed.py
pzhanggit: Done

allaffa (approved):

jychoi-hpc (approved):

hydragnn/models/Base.py
```diff
@@ -341,6 +348,8 @@ def __init__(self, input_dim, output_dim, num_nodes, hidden_dim_node, node_type)
                 denselayers.append(ReLU())
             denselayers.append(Linear(hidden_dim_node[-1], output_dim))
             self.mlp.append(Sequential(*denselayers))
+            if self.node_type == "mlp":
```
streeve, line 351: Let's switch to a variable loop bound instead

hydragnn/utils/distributed.py
streeve: `print_distributed` or remove

pzhanggit:
> Should we remove `get_model_or_module`? It seems like a good way to make a mistake again

Not sure. It is used also in the model saving. Probably related to the non-DDP mode we were discussing earlier. Let's keep it for now.

hydragnn/utils/distributed.py
pzhanggit: Done in 882bac3e57f5cafc07c3f6a69cc7efab35ee9cc7

hydragnn/models/Base.py
pzhanggit: Done in 882bac3e57f5cafc07c3f6a69cc7efab35ee9cc7

hydragnn/train/train_validate_test.py
```diff
@@ -216,89 +280,99 @@ def train(
...
+
     for data in iterate_tqdm(loader, verbosity):
         with record_function("zero_grad"):
             opt.zero_grad()
         with record_function("get_head_indices"):
             head_index = get_head_indices(model, data)
         with record_function("forward"):
             pred = model(data)
-            loss, tasks_loss = model.loss(pred, data.y, head_index)
+            loss, tasks_loss = model.module.loss(pred, data.y, head_index)
```
streeve, line 294: I think this means we always require DDP from now on - the non-DDP doesn't have `module` defined, right?

streeve (approved):
Not sure if we want to default to True for the reductions, but we can always change it later if it's an issue

hydragnn/train/train_validate_test.py
pzhanggit: Yes.

PR #142, opened 2022-07-08
removed check on maximum error from test_graph

allaffa:


allaffa:
@pzhanggit @jychoi-hpc 
Review PR#143 first. PR#143 solves some issues I had with pre-commit in order to push changes. 
Then, once we agree on that and we merge, I will rebase this PR and ask you to review it.

tests/test_graphs.py
```diff
@@ -122,18 +122,18 @@ def unittest_train_model(model_type, ci_input, use_lengths, overwrite_data=False
 
     # Set RMSE and sample MAE/max error thresholds
```
pzhanggit, line 123: Could you update the comment also?

pzhanggit (commented):
It looks good to me. One more to remove is the maximum sample error checking in https://github.com/ORNL/HydraGNN/blob/4468013857c1c1f07dfe4658b729c8e9d2533d8e/tests/test_model_loadpred.py#L40. Could you double check if we miss any other tests? Thanks.

tests/test_graphs.py
allaffa: Yep. Done

pzhanggit (approved):

PR #145, opened 2022-07-25
Update on using ADIOS and DDP for OGB PCQM4Mv2 and AISD HOMO-LUMO dataset

jychoi-hpc:
Sorry for a large PR. This is the update we made for the distributed training of OGB PCQM4Mv2 and AISD HOMO-LUMO dataset. Please feel free leave any questions and comments. 

The main update contains:
* ADIOS file genration
* data loading with ADIOS and DDP 
* Dataset class for OGB and CSCE.

Please merge with the "squash and merge" option when merging to clean up long commit history.


pzhanggit:
@jychoi-hpc  Thanks for the PR. It is quite complicated. A few general comments:
1. Could we keep only the critical files to demonstrate the examples and remove the others?
2. Could we also reduce the data files to keep some demonstration samples only? It is probably not a good idea to carry the large files around.

jychoi-hpc:
@pzhanggit  and @allaffa  CI test got an error since GCNN error is slightly larger than the threshold limit set here: https://github.com/jychoi-hpc/HydraGNN/blob/7ed9a9e5619c9eb3179dd2dc5da8eb8dfa9a593b/tests/test_graphs.py#L133

In my local machine GCNN got '0.157268' but the limit is '0.15'. Can we increase the threshold? I am also wondering if this error is deterministic or can be changing depending on randomness (e.g., depending on random sampling).

jychoi-hpc:
I have removed unnecessary files. I cannot figure out how to set seed perfectly. Just increased threshold for now.

pzhanggit (changes requested):
Thanks for the changes. I made a few more general comments, mostly about reorganizing/moving functions around and renaming.

examples/csce/ogb_utils.py
```diff
@@ -0,0 +1,209 @@
+import sys
```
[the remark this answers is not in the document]
jychoi-hpc: Make sense. I will move "orb_utils.py" in "utils" folder.

examples/ogb/ogb_utils.py
```diff
@@ -0,0 +1,301 @@
+import sys
```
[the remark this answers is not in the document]
jychoi-hpc: Will do the same thing.

hydragnn/train/train_validate_test.py
```diff
@@ -337,6 +346,13 @@ def train(
             for itask in range(len(tasks_loss)):
                 tasks_error[itask] += tasks_loss[itask] * data.num_graphs
 
+        ## (2022/05) jyc: Temporary fix - stop iteration after profile is done
```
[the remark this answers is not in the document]
jychoi-hpc: It's set to stop after profiling is done. I updated the comment.

hydragnn/models/Base.py
```diff
@@ -234,6 +236,7 @@ def _multihead(self):
             self.heads_NN.append(head_NN)
 
     def forward(self, data):
+        data = data.to(self.device)
```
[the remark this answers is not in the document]
jychoi-hpc: Got it. Will put in `train_validate_test.py`

hydragnn/utils/ogbdataset.py
```diff
@@ -0,0 +1,363 @@
+from mpi4py import MPI
```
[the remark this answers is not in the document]
jychoi-hpc: Make sense. I am renaming it.

hydragnn/utils/ogbdataset.py
```diff
@@ -0,0 +1,363 @@
...
+                    ad2.ConstantDims,
+                )
+                self.writer.Put(var, offset_arr, ad2.Mode.Sync)
+
+        self.writer.Close()
+        t1 = time.time()
+        log("Adios saving time (sec): ", (t1 - t0))
+
+
+class OGBDataset(torch.utils.data.Dataset):
```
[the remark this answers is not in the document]
jychoi-hpc: Make sense. I am renaming it as "AdiosDataset"

hydragnn/preprocess/load_data.py
```diff
@@ -171,7 +362,13 @@ def total_to_train_val_test_pkls(config):
         config["Dataset"]["path"][dataset_type] = (
             serialized_dir + "/" + serial_data_name
         )
-        if rank == 0:
+        if (not isdist) and (rank == 0):
+            with open(os.path.join(serialized_dir, serial_data_name), "wb") as f:
+                pickle.dump(minmax_node_feature, f)
+                pickle.dump(minmax_graph_feature, f)
+                pickle.dump(dataset, f)
+        elif isdist:
```
pzhanggit, line 370: If ... else ... switch?  We only need one to dump data.

pzhanggit: Or, we do not need the new changes?

hydragnn/preprocess/raw_dataset_loader.py
```diff
@@ -100,7 +127,19 @@ def load_raw_data(self):
...
 
-            for name in os.listdir(raw_data_path):
+            filelist = sorted(os.listdir(raw_data_path))
+            if self.dist:
+                ## Random shuffle filelist to avoid the same test/validation set
+                random.seed(43)
+                random.shuffle(filelist)
+
+                x = torch.tensor(len(filelist), requires_grad=False).to(get_device())
+                y = x.clone().detach().requires_grad_(False)
```
pzhanggit, line 137: Is `y` used to check that the same list is used on all ranks?

hydragnn/preprocess/raw_dataset_loader.py
```diff
@@ -356,6 +395,22 @@ def __normalize_dataset(self):
                         self.minmax_node_feature[1, ifeat],
                     )
                     n_index_start = n_index_end
+
+        ## Gather minmax in parallel
+        if self.dist:
```
pzhanggit, line 400: Thank you!

hydragnn/preprocess/raw_dataset_loader.py
```diff
@@ -100,7 +127,19 @@ def load_raw_data(self):
...
+                random.seed(43)
+                random.shuffle(filelist)
+
+                x = torch.tensor(len(filelist), requires_grad=False).to(get_device())
+                y = x.clone().detach().requires_grad_(False)
+                torch.distributed.all_reduce(x, op=torch.distributed.ReduceOp.MAX)
+                assert x == y
+                filelist = list(nsplit(filelist, self.world_size))[self.rank]
+                log("local filelist", len(filelist))
+            for name in filelist:
```
pzhanggit, line 142: If every rank now reads part of the data files, should we gather them before dumping? Or dump the data into multiple pickle files?

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,182 @@
...
+                print("unknown file name: ", row[0])
+                sys.exit(0)
+    return (
+        [trainsmiles, valsmiles, testsmiles],
+        [torch.tensor(trainset), torch.tensor(valset), torch.tensor(testset)],
+    )
+
+
+##################################################################################################################
+def get_node_attribute_name(types):
```
pzhanggit, line 87: Change it to return dimension also

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,182 @@
...
+
+    x1 = F.one_hot(torch.tensor(type_idx), num_classes=len(types))
+    x2 = (
+        torch.tensor([atomic_number, aromatic, sp, sp2, sp3, num_hs], dtype=torch.float)
+        .t()
+        .contiguous()
+    )
+
+    x = torch.cat([x1.to(torch.float), x2], dim=-1)
+    # x = torch.tensor([atomic_number], dtype=torch.float).view(-1,1)
```
pzhanggit, line 162: Comment line 162, 163, 167, 180, and 182 are not needed any more. Let's remove them

pzhanggit (commented):
More comments @jychoi-hpc  Some of the comments are applicable to multiple examples. Could you check them when modifying? Thanks. Also, let me know if any comments unclear.

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    # print(x)
+
+    y = ytarget  # .squeeze()
+
+    # data = Data(x=x, z=z, edge_index=edge_index, edge_attr=edge_attr, y=y)
+    data = Data(x=x, edge_index=edge_index, edge_attr=edge_attr, y=y)
+    if var_config is not None:
+        graph_features_dim = [1]
+        node_feature_dim = [1]
```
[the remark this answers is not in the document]
jychoi-hpc: Good idea. Then, we add "graph_feature_dims" and "intput_node_feature_dims" in json file, right?

examples/ising_model/ising_model.json
```diff
@@ -16,7 +16,8 @@
             "dim": [1],
             "column_index": [0]
         },
-        "compositional_stratified_splitting": 0
+        "compositional_stratified_splitting": 0,
```
allaffa, line 19: I know that 0 and 1 are treated as `False` and `True`. But maybe it is better to actually use the boolean values?

examples/ising_model/ising_model.json
```diff
@@ -16,7 +16,8 @@
...
+        "rotational_invariance": 0
```
allaffa, line 20: I know that 0 and 1 are treated as False and True. But maybe it is better to actually use the boolean values?

examples/ising_model/ising_model.json
```diff
@@ -38,7 +39,8 @@
                     "type": "mlp"
                 }
             },
-            "task_weights": [1.0, 1.0]
+            "task_weights": [1.0, 1.0],
+            "periodic_boundary_conditions": 0
```
allaffa, line 43: I know that 0 and 1 are treated as False and True. But maybe it is better to actually use the boolean values?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+                total_energy, atomic_features = E_dimensionless(
+                    config, L, spin_function, scale_spin
+                )
+
+                write_to_file(total_energy, atomic_features, count_config, dir, prefix)
+
+                count_config = count_config + 1
+
+
+class AdioGGO:
```
allaffa, line 107: I think this is a typo. Shouldn't it be `AdiosGGO`?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+from hydragnn.utils.model import print_model
+
+import numpy as np
+import adios2 as ad2
+
+import torch_geometric.data
+import torch
+import torch.distributed as dist
+
+torch.multiprocessing.set_start_method("fork", force=True)
```
allaffa, line 28: Why do we need this?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
+import os, json
+import matplotlib.pyplot as plt
+
+import logging
+import sys
+from tqdm import tqdm
+import mpi4py
```
allaffa, line 7: Do we need this?

hydragnn/utils/ogb_utils.py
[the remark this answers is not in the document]
pzhanggit: Yep, we can add them either in the json file or in the script (train_gap.py).

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+from mpi4py import MPI
```
allaffa, line 8: We tried as much as possible to avoid explicit use of MPI directives so far. 
Is there no way to replace the explicit call to MPI with equivalent calls that solely rely on `torch.distributed `?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+                data = self.dataset[label][0]
+                self.io.DefineAttribute("%s/keys" % label, data.keys)
+                keys = sorted(data.keys)
+
+            for k in keys:
+                arr_list = [data[k].cpu().numpy() for data in self.dataset[label]]
+                m0 = np.min([x.shape for x in arr_list], axis=0)
+                m1 = np.max([x.shape for x in arr_list], axis=0)
+                wh = np.where(m0 != m1)[0]
+                assert len(wh) < 2
```
allaffa, line 156: why do you need this ?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+            t2 = time.time()
+            info("Adios reading time (sec): ", (t2 - t0))
+
+        t1 = time.time()
+        info("Data loading time (sec): ", (t1 - t0))
+
+    def download(self):
+        path = download_url(self.url, self.prefix)
+        extract_tar(path, self.prefix)
+        # os.unlink(path)
```
allaffa, line 284: If never used, maybe we could remove this?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+
+        self.dataset = dict()
+        self.attributes = dict()
+        self.adios = ad2.ADIOS()
+        self.io = self.adios.DeclareIO(self.filename)
+
+    def add_global(self, vname, arr):
+        self.attributes[vname] = arr
+
+    def add(self, label, data: torch_geometric.data.Data):
```
allaffa, line 122: I assume that here you use `data: torch_geometric.data.Data` to suggest what should be the type of the object. However, inside the body of this function you also check if `data` is a list. Is ti really useful to set `data: torch_geometric.data.Data`  then ?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+                self.io.DefineAttribute("%s/keys" % label, data.keys)
+                keys = sorted(data.keys)
+
+            for k in keys:
+                arr_list = [data[k].cpu().numpy() for data in self.dataset[label]]
+                m0 = np.min([x.shape for x in arr_list], axis=0)
+                m1 = np.max([x.shape for x in arr_list], axis=0)
+                wh = np.where(m0 != m1)[0]
+                assert len(wh) < 2
+                vdim = wh[0] if len(wh) == 1 else len(arr_list[0].shape) - 1
```
allaffa, line 157: In line 155 you already define `wh` by accessing the first entry of the numpy array defined by `np.where(m0 != m1)`. Moreover, in line 156 you have an assert that makes the code automatically crash if `wh` has more than one entry. 
Shouldn't `wh[0]` be equal to `wh` itself ?

allaffa (commented):
I added comments and questions to see if I understand correctly what's going on in the code.

examples/csce/train_gap.py
```diff
@@ -0,0 +1,514 @@
...
+    ##################################################################################################################
+    inputfilesubstr = args.inputfilesubstr
+    input_filename = os.path.join(dirpwd, "csce_" + inputfilesubstr + ".json")
+    ##################################################################################################################
+    # Configurable run choices (JSON file that accompanies this example script).
+    with open(input_filename, "r") as f:
+        config = json.load(f)
+    verbosity = config["Verbosity"]["level"]
+    var_config = config["NeuralNetwork"]["Variables_of_interest"]
+    var_config["output_names"] = [
```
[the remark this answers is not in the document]
jychoi-hpc: Now "graph_feature_dims" is added in the json file (csce_gap.json and orb_gap.json)

hydragnn/preprocess/load_data.py
```diff
@@ -30,6 +31,170 @@
...
+    for block in re.findall(r"({[\d,:]+})", envstr):
+        start, cnt = list(map(int, re.findall(r"\d+", block)))
+        for i in range(start, start + cnt):
+            plist.append(i)
+    return plist
+
+
+class HydraSimpleDataLoader(DataLoader):
+    def __init__(self, dataset, **kwargs):
+        super(HydraDataLoader, self).__init__(dataset, **kwargs)
```
allaffa, line 65: Shouldn't this be `HydraSimpleDataLoader` ?

allaffa: Do you ever use this simplified data loader? Why do you have two Hydra data loaders?

hydragnn/preprocess/load_data.py
```diff
@@ -171,7 +362,13 @@ def total_to_train_val_test_pkls(config):
...
-        if rank == 0:
+        if (not isdist) and (rank == 0):
+            with open(os.path.join(serialized_dir, serial_data_name), "wb") as f:
+                pickle.dump(minmax_node_feature, f)
+                pickle.dump(minmax_graph_feature, f)
+                pickle.dump(dataset, f)
+        elif isdist:
+            ## parallel processing. Everyone need to read own total pickle data
             with open(os.path.join(serialized_dir, serial_data_name), "wb") as f:
                 pickle.dump(minmax_node_feature, f)
```
allaffa, line 373: if the data is distributed and multiple processes executed this "dumping" (which I think is a writing), is it possible that this line causes a race condition?

examples/ising_model/train_ising.py
```diff
@@ -0,0 +1,502 @@
...
+        self.comm = comm
+        self.rank = comm.Get_rank()
+
+        self.data_object = dict()
+        info("Adios reading:", self.filename)
+
+        if self.rank == 0:
+            if not os.path.exists(filename) and self.url is not None:
+                self.prefix = os.path.dirname(self.filename)
+                self.download()
```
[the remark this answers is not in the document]
jychoi-hpc: It was experimental. I am removing now.

examples/ising_model/train_ising.py
jychoi-hpc: It was experimental. I will remove

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+def gapfromsmiles(smilestr, model):
+    ## gap_true can be replaced by random numbers when use
+    gap_rand = 0.0
+    data_graph = generate_graphdata(smilestr, gap_rand)
+    pred = model(data_graph)
+    return pred[0][0].item()
+
+
```
pzhanggit, line 111:
```suggestion
```

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,178 @@
...
+        csvreader = csv.reader(file)
+        print(next(csvreader))
+        for row in csvreader:
+            smiles.append(row[0])
+            strset.append(row[1])
+            yvals.append(float(row[-1]))
+    return smiles, yvals, strset
+
+
+def datasets_load(datafile, sampling=None, seed=None):
```
pzhanggit, line 50: Move this function to ogb/train_gap.py ? to be consistent with csce

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,178 @@
...
+    strset = []
+    yvals = []
+    with open(datafile, "r") as file:
+        csvreader = csv.reader(file)
+        print(next(csvreader))
+        for row in csvreader:
+            smiles.append(row[0])
+            strset.append(row[1])
+            yvals.append(float(row[-1]))
+    return smiles, yvals, strset
```
pzhanggit, line 47:
```suggestion
```

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,178 @@
...
+
+def get_trainset_stat(filedir):
+    with open(filedir, "rb") as f:
+        max_feature = pickle.load(f)
+        min_feature = pickle.load(f)
+        mean_feature = pickle.load(f)
+        std_feature = pickle.load(f)
+    return max_feature, min_feature, mean_feature, std_feature
+
+
```
pzhanggit, line 35:
```suggestion
```

hydragnn/utils/ogb_utils.py
```diff
@@ -0,0 +1,178 @@
+import sys
```
pzhanggit, line 1: Rename this file to be something like smiles_utils.py?

pzhanggit (commented):
More minor comments... The comments are mostly about removing unused lines and functions.

examples/csce/train_gap.py
```diff
@@ -0,0 +1,417 @@
...
+    with open(datafile, "r") as file:
+        csvreader = csv.reader(file)
+        print(next(csvreader))
+        for row in csvreader:
+            if (sampling is not None) and (random.random() > sampling):
+                continue
+            smiles_all.append(row[1])
+            values_all.append([float(row[-2])])
+    print("Total:", len(smiles_all), len(values_all))
+    atomtypes_dict = {}
```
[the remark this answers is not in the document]
jychoi-hpc: Great. I will remove.

examples/csce/train_gap.py
```diff
@@ -0,0 +1,417 @@
...
+        create_plots=True,
+    )
+
+    hydragnn.utils.save_model(model, optimizer, log_name)
+    hydragnn.utils.print_timers(verbosity)
+
+    if args.mae:
+        ##################################################################################################################
+        fig, axs = plt.subplots(1, 3, figsize=(18, 6))
+        isub = -1
```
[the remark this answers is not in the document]
jychoi-hpc: Will remove.

examples/ogb/train_gap.py
```diff
@@ -0,0 +1,387 @@
...
+        info([len(x) for x in values_sets])
+        dataset_lists = [[] for dataset in values_sets]
+        for idataset, (smileset, valueset) in enumerate(zip(smiles_sets, values_sets)):
+            if norm_yflag:
+                valueset = (
+                    valueset - torch.tensor(var_config["ymean"])
+                ) / torch.tensor(var_config["ystd"])
+                print(valueset[:, 0].mean(), valueset[:, 0].std())
+                print(valueset[:, 1].mean(), valueset[:, 1].std())
+                print(valueset[:, 2].mean(), valueset[:, 2].std())
```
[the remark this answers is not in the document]
allaffa: @pzhanggit 
The command lines with your suggestions seem to be exactly the same as the ones that are already written in the code. Maybe @jychoi-hpc already made your suggested changes ?

hydragnn/utils/print_utils.py
```diff
@@ -53,9 +53,9 @@ def print_distributed(verbosity_level, *args):
     return print_verbose(*args)
 
 
-def iterate_tqdm(iterator, verbosity_level):
+def iterate_tqdm(iterator, verbosity_level, *args, **kwargs):
```
allaffa, line 56: what is the advantaged of having *args and **kwargs as additional arguments for `iterate_tqdm` ?

allaffa (commented):
I added a few more questions

examples/csce/train_gap.py
```diff
@@ -0,0 +1,417 @@
...
+                if args.format == "pickle":
+                    fname = "examples/csce/dataset/pickle/csce_gap-%s-%d.pk" % (
+                        setname[idataset],
+                        rx.start + i,
+                    )
+                    with open(fname, "wb") as f:
+                        pickle.dump(data, f)
+
+        ## local data
+        _trainset = dataset_lists[0]
```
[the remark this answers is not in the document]
jychoi-hpc: Yes, I will add.

examples/csce/train_gap.py
```diff
@@ -0,0 +1,417 @@
...
+    elif args.format == "csv":
+        fact = CSCEDatasetFactory(
+            "examples/csce/dataset/csce_gap_synth.csv",
+            args.sampling,
+            var_config=var_config,
+        )
+        trainset = CSCEDataset(fact, "trainset")
+        valset = CSCEDataset(fact, "valset")
+        testset = CSCEDataset(fact, "testset")
+    else:
```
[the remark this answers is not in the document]
jychoi-hpc: Great. Thank you for the check I have added "pickle" read.

hydragnn/preprocess/utils.py
```diff
@@ -16,17 +16,56 @@
...
     for loader in [train_loader, val_loader, test_loader]:
         for data in loader.dataset:
             nodes_num_list.append(data.num_nodes)
+        if len(list(set(nodes_num_list))) > 1:
+            graph_size_variable = True
+            return graph_size_variable
+    return graph_size_variable
+
+
+def check_if_graph_size_variable_mpi(train_loader, val_loader, test_loader):
```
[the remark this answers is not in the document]
jychoi-hpc: I will remove `check_if_graph_size_variable_mpi` as well as `calculate_PNA_degree_mpi` as the MPI versions will not be used anymore.

hydragnn/train/train_validate_test.py
```diff
@@ -314,17 +321,20 @@ def train(
...
         profiler = Profiler()
 
-    total_error = 0
-    tasks_error = torch.zeros(model.module.num_heads, device=loader.dataset[0].y.device)
+    total_error = torch.tensor(0.0, device=get_device())
+    tasks_error = torch.zeros(model.module.num_heads, device=get_device())
     num_samples_local = 0
     model.train()
+    if isinstance(loader, HydraDataLoader):
+        loader.device = get_device()
```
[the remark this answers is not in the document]
jychoi-hpc: I will remove them. It was an experimental feature.

examples/csce/train_gap.py
```diff
@@ -0,0 +1,417 @@
...
+        raise NotImplementedError("No supported format: %s" % (args.format))
+
+    info("Adios load")
+    info(
+        "trainset,valset,testset size: %d %d %d"
+        % (len(trainset), len(valset), len(testset))
+    )
+
+    (train_loader, val_loader, test_loader,) = hydragnn.preprocess.create_dataloaders(
```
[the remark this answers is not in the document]
jychoi-hpc: It is correct. DistributedSampler will do subset for each process.

examples/ogb/train_gap.py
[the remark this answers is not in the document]
jychoi-hpc: Not sure when I did. Anyway, it is gone now.

hydragnn/utils/ogb_utils.py
jychoi-hpc: Great. I will remove.

hydragnn/utils/ogb_utils.py
jychoi-hpc: Thanks. I will remove.

hydragnn/utils/ogb_utils.py
jychoi-hpc: Great. I will remove.

hydragnn/utils/ogb_utils.py
jychoi-hpc: I think this is a good suggestion. I will rename it.

hydragnn/utils/print_utils.py
jychoi-hpc: tqdm has many options we can use. We can pass such options with "*args, **kwargs" (This is also optional parameters. It will not break any previous usage).  

For example, I have added "desc" option in `tqdm`s in `train_validate_test.py`:
```
        for data in iterate_tqdm(
            test_loader, verbosity, desc="Collecting node feature"
        ):
```

jychoi-hpc (commented):
Thank you for the detailed check. I think I have addressed all the comments. Feel free to check and leave further comments.

examples/ogb/train_gap.py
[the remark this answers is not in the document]
pzhanggit: > 

Not following exactly what you meant @allaffa . The lines (that were suggested to be removed) were there in this round of review.

examples/csce/train_gap.py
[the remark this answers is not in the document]
pzhanggit: Then do we need to gather loaded samples after each process's reading for example in `SimplePickleDataset` so that every process have all the samples?

hydragnn/preprocess/utils.py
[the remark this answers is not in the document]
pzhanggit: This one is missing?

examples/csce/train_gap.py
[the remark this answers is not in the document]
jychoi-hpc: SimplePickleDataset is created to compare Adios reading vs single pickle file per graph reading in the paper (ie, number of graphs == number of pickle file). So, no need to gather is needed here. If this is confusing, I can remove since this feature is added for the paper writing.

hydragnn/preprocess/utils.py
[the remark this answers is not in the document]
jychoi-hpc: Sorry. Not clear to me what you are asking. Do you mean you cannot find "check_if_graph_size_variable_mpi"? I have deleted it in favor of using "check_if_graph_size_variable_dist".

hydragnn/utils/ogb_utils.py
jychoi-hpc: Good point. I agree with you. I will move to `ogb/train_gap.py`

hydragnn/preprocess/utils.py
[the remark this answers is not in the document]
pzhanggit: The opposite. I meant, the functions are still there in the script and need to be removed.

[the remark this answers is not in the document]
pzhanggit: Are they not showing up on your side?

examples/csce/train_gap.py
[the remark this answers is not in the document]
pzhanggit: I see. Somehow I was confused and thought that the number of pickles files is equal to the number of processes and each process will only read the subset in the corresponding pickle file. Now it is clear. No need to remove. Thanks for the explanation.

hydragnn/preprocess/utils.py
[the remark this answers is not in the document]
jychoi-hpc: I see. I have removed and don't see the function "check_if_graph_size_variable_mpi" anymore from my side.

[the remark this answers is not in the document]
pzhanggit: Quite strange, I can still see the function here, https://github.com/jychoi-hpc/HydraGNN/blob/bd80d1027895865dfaa4c9874cfa6fb7a7b6ce85/hydragnn/preprocess/utils.py#L33.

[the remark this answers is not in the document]
jychoi-hpc: I am sorry. My fault. I haven't saved the file. The text tool I am using is quite good at maintaining cache :) I just pushed.

hydragnn/preprocess/raw_dataset_loader.py
jychoi-hpc: It makes sense more to use the old way. I just revert the change back to the previous one. I.e., setting `config["Dataset"]["name"]` in `train_ising.py`. Thank you for double-checking.

hydragnn/utils/adiosdataset.py
```diff
@@ -0,0 +1,389 @@
...
+        if os.getenv("SLURM_NTASKS_PER_NODE"):
+            ## Perlmutter
+            self.nrank_per_node = int(os.environ["SLURM_NTASKS_PER_NODE"])
+
+        self.local_rank = self.rank % self.nrank_per_node
+        log("local_rank", self.local_rank)
+
+        self.data = dict()
+        self.preload = preload
+        self.preflight = False
```
allaffa, line 171: @jychoi-hpc 
What is the utility of preflight? Is it used for pre-fetching future data and store it in cache?

examples/csce/train_gap.py
```diff
@@ -0,0 +1,414 @@
...
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("inputfilesubstr", help="input file substr")
+    parser.add_argument("--sampling", type=float, help="sampling ratio", default=None)
+    parser.add_argument(
+        "--preonly",
+        action="store_true",
+        help="preprocess only. Adios saving and no train",
```
allaffa, line 149: @jychoi-hpc 
Actually, if args.preonly == True, you also allow to write the data on .pkl files. 
Am I right? If this is the case, then "Adios saving and no train" is confusing

allaffa (commented):
I added a couple of more comments

hydragnn/utils/adiosdataset.py
jychoi-hpc: I am sorry for the lack of comment.  Returning empty object is intentional. It is an experimental feature I made for the paper. I named as a 'preflight'. 
It will collect only index of data to load and "populate" call will load all the data data from disk. I was testing this for the paper.

I put more comments for the options available in `AdiosDataset`.

examples/csce/train_gap.py
jychoi-hpc: Agree. I will change as follows:
```
    parser.add_argument(
        "--preonly",
        action="store_true",
        help="preprocess only (no training)",
    )
```

examples/ising_model/train_ising.py
allaffa: In line 155 you are extracting the first element oof the bumpy array. 
If this is the case, should it never be that `len(wh)=1`

hydragnn/utils/adiosdataset.py
```diff
@@ -0,0 +1,456 @@
...
+        label: str
+            data label to load, such as trainset, testing, and valset
+        comm: MPI_comm
+            MPI communicator
+        preload: bool, optional
+            Option to preload all the dataset into a memory
+        shmem: bool, optional
+            Option to use shmem to share data between processes in the same node
+        enable_cache: bool, optional
+            Opption to cache data object which was already read
```
allaffa, line 198:
```suggestion
            Opption to cache data object which was already read
```

allaffa: @jychoi-hpc 

I know I am very picky here, but there is a typo :)

allaffa (commented):
New comments

hydragnn/utils/adiosdataset.py
```diff
@@ -0,0 +1,456 @@
...
+                data = self.dataset[label][0]
+                self.io.DefineAttribute("%s/keys" % label, data.keys)
+                keys = sorted(data.keys)
+
+            for k in keys:
+                arr_list = [data[k].cpu().numpy() for data in self.dataset[label]]
+                m0 = np.min([x.shape for x in arr_list], axis=0)
+                m1 = np.max([x.shape for x in arr_list], axis=0)
+                wh = np.where(m0 != m1)[0]
+                assert len(wh) < 2
```
allaffa, line 102: @jychoi-hpc 
Sorry but I still struggle to understand why this assert is needed. 
In the previous line, you explicitly extract the first entry of the bumpy array. 
Once you do that, shouldn't it be alway true that `len(wh)=1` ?

allaffa (commented):
New comments added.

pzhanggit (commented):
Add two more comments...
It looks fine to me now. I have verified the codes run fine on Summit for examples: csce, ising, ogb, and lsms. As soon as the current comments are resolved, I will approve the PR.

hydragnn/utils/adiosdataset.py
jychoi-hpc: Not a problem. `np.where` returns a list of arrays:
```
>>> m0 = np.array([6,4])
>>> m1 = np.array([100,4])
>>> np.where(m0 != m1)
(array([0]),)
```
`np.where(m0 != m1)**[0]**` extracts the array of matching indices. 
Then, `len(wh)` can be either 0 (for scalar) or 1 (for 1d shape vector)

hydragnn/utils/distributed.py
```diff
@@ -149,9 +154,8 @@ def setup_ddp():
         )
 
         if not dist.is_initialized():
-            dist.init_process_group(
-                backend=backend, rank=int(world_rank), world_size=int(world_size)
-            )
+            dist.init_process_group(backend=backend, init_method="env://")
```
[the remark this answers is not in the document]
jychoi-hpc: I thought "env://` is the default way in pytorch. It looks like both are same:
https://pytorch.org/docs/stable/distributed.html#torch.distributed.init_process_group

examples/csce/train_gap.py
```diff
@@ -0,0 +1,414 @@
...
+            ax.text(
+                minv + 0.1 * (maxv - minv),
+                maxv - 0.1 * (maxv - minv),
+                "MAE: {:.2f}".format(error_mae),
+            )
+        if rank == 0:
+            fig.savefig("./logs/" + log_name + "/" + varname + "_all.png")
+        plt.close()
+
+    if not args.noadios:
```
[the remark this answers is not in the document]
jychoi-hpc: Nice finding. It's the error. I will fix.

pzhanggit (approved):
Thanks for fixing all my comments. It looks good to me now.

allaffa (approved):

PR #147, opened 2022-09-12
Lsms example

pzhanggit:
Update lsms  following the style in ising example

### DIFF
--- a/examples/lsms/lsms.py
+++ b/examples/lsms/lsms.py
@@ -1,5 +1,202 @@
-import os
-import hydragnn
+import os, json
+import logging
+import sys
+from mpi4py import MPI
+import argparse
 
-filepath = os.path.join(os.path.dirname(__file__), "lsms.json")
-hydragnn.run_training(filepath)
+import hydragnn
+from hydragnn.utils.time_utils import Timer
+from hydragnn.utils.config_utils import get_log_name_config
+from hydragnn.preprocess.raw_dataset_loader import RawDataLoader
+from hydragnn.utils.model import print_model
+
+import torch
+import torch.distributed as dist
+
+
+def info(*args, logtype="info", sep=" "):
+    getattr(logging, logtype)(sep.join(map(str, args)))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--preonly",
+        action="store_true",
+        help="preprocess only. Adios or pickle saving and no train",
+    )
+    parser.add_argument("--inputfile", help="input file", type=str, default="lsms.json")
+    group = parser.add_mutually_exclusive_group()
+    group.add_argument(
+        "--adios",
+        help="Adios dataset",
+        action="store_const",
+        dest="format",
+        const="adios",
+    )
+    group.add_argument(
+        "--pickle",
+        help="Pickle dataset",
+        action="store_const",
+        dest="format",
+        const="pickle",
+    )
+    parser.set_defaults(format="pickle")
+
+    args = parser.parse_args()
+
+    dirpwd = os.path.dirname(__file__)
+    input_filename = os.path.join(dirpwd, args.inputfile)
+    with open(input_filename, "r") as f:
+        config = json.load(f)
+    hydragnn.utils.setup_log(get_log_name_config(config))
+    ##################################################################################################################
+    # Always initialize for multi-rank training.
+    comm_size, rank = hydragnn.utils.setup_ddp()
+    ##################################################################################################################
+    comm = MPI.COMM_WORLD
+    ## Set up logging
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%%(levelname)s (rank %d): %%(message)s" % (rank),
+        datefmt="%H:%M:%S",
+    )
+
+    os.environ["SERIALIZED_DATA_PATH"] = dirpwd + "/dataset"
+    datasetname = config["Dataset"]["name"]
+    fname_adios = dirpwd + "/dataset/%s.bp" % (datasetname)
+    config["Dataset"]["name"] = "%s_%d" % (datasetname, rank)
+    if args.preonly:
+        for dataset_type, raw_data_path in config["Dataset"]["path"].items():
+            if not os.path.isabs(raw_data_path):
+                raw_data_path = os.path.join(dirpwd, raw_data_path)
+            if not os.path.exists(raw_data_path):
+                raise ValueError("Folder not found: ", raw_data_path)
+            config["Dataset"]["path"][dataset_type] = raw_data_path
+
+        ## each process saves its own data file
+        loader = RawDataLoader(config["Dataset"], dist=True)
+        loader.load_raw_data()
+
+        ## Read total pkl and split (no graph object conversion)
+        hydragnn.preprocess.total_to_train_val_test_pkls(config, isdist=True)
+
+        ## Read each pkl and graph object conversion with max-edge normalization
+        (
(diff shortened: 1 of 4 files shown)


examples/lsms/lsms.py
```diff
@@ -1,5 +1,202 @@
...
+            adwriter.add("testset", testset)
+            adwriter.add_global("minmax_node_feature", loader.minmax_node_feature)
+            adwriter.add_global("minmax_graph_feature", loader.minmax_graph_feature)
+            adwriter.save()
+        sys.exit(0)
+
+    timer = Timer("load_data")
+    timer.start()
+    if args.format == "adios":
+        import adios2 as ad2
```
jychoi-hpc, line 107: You can remove this line. `AdiosDataset` will handle it.

examples/lsms/lsms.py
```diff
@@ -1,5 +1,202 @@
...
+
+        ## Read each pkl and graph object conversion with max-edge normalization
+        (
+            trainset,
+            valset,
+            testset,
+        ) = hydragnn.preprocess.load_data.load_train_val_test_sets(config, isdist=True)
+
+        if args.format == "adios":
+            import adios2 as ad2
```
jychoi-hpc, line 92: You can remove this line. AdiosDataset will handle it.

hydragnn/train/train_validate_test.py
```diff
@@ -67,7 +67,7 @@ def train_validate_test(
         nodes_num_list = []
         ## (2022/05) : FIXME: using test_loader.datast caused a bottleneck for large data
         for data in iterate_tqdm(
-            test_loader, verbosity, desc="Collecting node feature"
+            test_loader.dataset, verbosity, desc="Collecting node feature"
```
jychoi-hpc, line 70: You can remove `.dataset` and just use `test_loader`. Using `test_loader.dataset` will slow down for other cases with larger dataset (e.g., OGB, CSCE).

jychoi-hpc (changes requested):
I just made a few comments on cleaning up the code and using data loader.

hydragnn/train/train_validate_test.py
pzhanggit: We aim to collect node information, e.g., number of nodes for each graph, here for visualization. If we remove `.dataset`, the information we get is batch info, e.g., total number of nodes of the entire batch (=num_nodes*batch_size for constant graph size), instead of for each graph. Yeah, it can be slow for larger datasets. Any suggestions to improve it?

jychoi-hpc: After the conversation, you can leave as it is. It makes sense to me.

jychoi-hpc (approved):

examples/lsms/lsms.py
```diff
@@ -1,5 +1,200 @@
...
+
+
+def info(*args, logtype="info", sep=" "):
+    getattr(logging, logtype)(sep.join(map(str, args)))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--preonly",
```
allaffa, line 24: Unless the flag `--preonly` is explicitly called when `lsms.py` is launched, the conversion of raw data into .pkl files is never executed. 
If one starts with just raw data files and launches `lsms.py`, the code crashes because the code skips the creation of the .pkl files, searches for the .pkl files assuming that they already exists, and crashes because it does no find them.

allaffa: But I like the idea of providing the options of only converting the raw data into either .pkl or adios without necessarily kicking off the training.

examples/lsms/lsms.py
```diff
@@ -1,5 +1,200 @@
...
+        level=logging.INFO,
+        format="%%(levelname)s (rank %d): %%(message)s" % (rank),
+        datefmt="%H:%M:%S",
+    )
+
+    os.environ["SERIALIZED_DATA_PATH"] = dirpwd + "/dataset"
+    datasetname = config["Dataset"]["name"]
+    fname_adios = dirpwd + "/dataset/%s.bp" % (datasetname)
+    config["Dataset"]["name"] = "%s_%d" % (datasetname, rank)
+    if args.preonly:
```
allaffa, line 69: Unless the flag --preonly is explicitly called when lsms.py is launched, the conversion of raw data into .pkl files is never executed.
If one starts with just raw data files and launches lsms.py, the code crashes because the code skips the creation of the .pkl files, searches for the .pkl files assuming that they already exists, and crashes because it does no find them.

examples/lsms/lsms.py
```diff
@@ -1,5 +1,200 @@
...
+        ## Read total pkl and split (no graph object conversion)
+        hydragnn.preprocess.total_to_train_val_test_pkls(config, isdist=True)
+
+        ## Read each pkl and graph object conversion with max-edge normalization
+        (
+            trainset,
+            valset,
+            testset,
+        ) = hydragnn.preprocess.load_data.load_train_val_test_sets(config, isdist=True)
+
```
allaffa, line 90: I think the the if condition at line 69 should me removed. 
Instead, here you should say 

```suggestion
if args.preonly:
    break
```

hydragnn/utils/config_utils.py
```diff
@@ -201,7 +201,7 @@ def get_log_name_config(config):
         + "-bs-"
         + str(config["NeuralNetwork"]["Training"]["batch_size"])
         + "-data-"
-        + config["Dataset"]["name"]
+        + config["Dataset"]["name"][: config["Dataset"]["name"].rindex("_")]
```
allaffa, line 204: This assumes that the data name always contain "_" in it. 
Do we really need this?

allaffa (changes requested):
First round of review done

examples/lsms/lsms.py
pzhanggit: Good point. I will make update the `--preonly` part. Was following the same style in other examples.

hydragnn/utils/config_utils.py
pzhanggit: > This assumes that the data name always contain "_" in it. 
---Good point, I should have used `rfind`. Will update this part.
Do we really need this?
---Yes, because the `config["Dataset"]["name"] = "%s_%d" % (datasetname, rank)` and we do not want `rank` show up  in output folder name.

examples/lsms/lsms.py
pzhanggit: We do not want to remove line 69 because we may want to load existing pickle/adios data with train/test/val splits for reproducibility or continuous training. 
In e6c26ede419d1bad68652df18b7a05793f894c65, I add a new flag `loadexistingsplit` and make it default to do raw data conversion&splitting unless specified. Let me know what you think.

allaffa (approved):

allaffa:
@pzhanggit I just re-ran the `lsms.py` Python script on debug mode on my laptop. 
Seems good to me.

hydragnn/utils/config_utils.py
allaffa: @pzhanggit If you run the code in sequential and config["Dataset"]["name"] doe snot contain a "_", the code crashes.

allaffa (commented):
@pzhanggit one more comment on "_"

allaffa (changes requested):
One more comment on "_"

allaffa (approved):

allaffa:
@pzhanggit sorry for the confusion. I saw the last change on the "_". I think this PR is ready to merge

PR #148, opened 2022-09-15
use_zero_redundancy flag added to enable ZeroRedundancyOptimizer

allaffa:


hydragnn/utils/optimizer.py
```diff
@@ -37,3 +38,76 @@ def select_optimizer(model, config):
...
+        raise NameError("The string used to identify the optimizer is NOT recognized")
+
+    return optimizer
+
+
+def select_optimizer(model, config):
+    use_zero = False
+
+    if "use_zero" in config["optimizer"]:
+        use_zero = config["optimizer"]["use_zero"]
```
pzhanggit, line 108: It should be `config['Training']["optimizer"]` here, to be consistent with `ci.json` input and` test_optimizer.py`. Currently `ZeroRedundancyOptimizer` will not be called.

pzhanggit (changes requested):

hydragnn/utils/optimizer.py
```diff
@@ -37,3 +38,76 @@ def select_optimizer(model, config):
...
+    else:
+        raise NameError("The string used to identify the optimizer is NOT recognized")
+
+    return optimizer
+
+
+def select_optimizer(model, config):
+    use_zero = False
+
+    if "use_zero" in config["optimizer"]:
```
allaffa, line 107:
```suggestion
    if "use_zero" in config['Training']["optimizer"]:
```

hydragnn/utils/optimizer.py
allaffa, line 108:
```suggestion
        use_zero = config['Training']["optimizer"]["use_zero"]
```

allaffa: @pzhanggit sorry but I do not agree. the `config` passed in input to the routines here is not the entire config. It is just the section associated with the Training.

pzhanggit: I meant config['Training']["use_zero"].

pzhanggit: Okay, after checking line 81 of `/hydragnn/run_training.py`, it should be `config` and `config["use_zero"]`, because `["use_zero"]` is not defined as item in `["optimizer"]`?

allaffa: Yes, that is correct. I was thinking whether I should define `["use_zero"]` as item in `["optimizer"]` or not. 
`ZeroRedundancyOptimizer` is a wrapper that applies to all optimizers. What would you do?

pzhanggit: Either one works for me. Putting ["use_zero"] as item in ["optimizer"] is neater but requires more changes to the code.

allaffa: If I were to create a new subsection in the configure file, both ["use_zero"] and ["learning_rate"] would go under ["optimizer"], whereas everything else would remain where it already is. Correct?

allaffa:
@pzhanggit @jychoi-hpc  This PR is ready for another round.

hydragnn/utils/optimizer.py
```diff
@@ -8,32 +9,119 @@
...
     else:
         raise NameError("The string used to identify the optimizer is NOT recognized")
 
     return optimizer
+
+
+def select_optimizer(model, config):
+    use_zero = False
+
+    if "use_zero_redundancy" in config["Optimizer"]["type"]:
```
pzhanggit, line 121: Typo? 
```suggestion
    if "use_zero_redundancy" in config["Optimizer"]:
```

hydragnn/utils/optimizer.py
```diff
@@ -8,32 +9,119 @@
...
-    elif config["optimizer"] == "Adagrad":
-        optimizer = torch.optim.Adagrad(model.parameters(), lr=config["learning_rate"])
-    elif config["optimizer"] == "Adamax":
-        optimizer = torch.optim.Adamax(model.parameters(), lr=config["learning_rate"])
-    elif config["optimizer"] == "AdamW":
-        optimizer = torch.optim.AdamW(model.parameters(), lr=config["learning_rate"])
-    elif config["optimizer"] == "RMSprop":
-        optimizer = torch.optim.RMSprop(model.parameters(), lr=config["learning_rate"])
-    elif config["optimizer"] == "FusedLAMB":
+    if config["Optimizer"]["type"] == "SGD":
```
pzhanggit, line 15: Since now all the Optimizer related parameters have been put inside `config["Optimizer"]`, we can pass `config["Optimizer"]` to `select_optimizer` to get rid of all the `["Optimizer"]` in `optimizer.py`.

jychoi-hpc (approved):
Great. This looks good to me. I tested on Summit with the OGB case and it went ok.

pzhanggit (approved):

PR #152, opened 2022-10-13
fix head indexing bug 

pzhanggit:
Currently, when there are multiple outputs/heads and all of them are graph-level properties, only the first property is read repeatedly due to indexing bug.


jychoi-hpc (approved):
I haven't tested by myself. But, it looks ok to me.

allaffa (approved):

PR #156, opened 2022-11-16
Add atomic descriptors from Mendeleev

pzhanggit:
Add atomic descriptors from Mendeleev (https://mendeleev.readthedocs.io/en/stable/index.html)

### DIFF
--- a/hydragnn/utils/atomicdescriptors.py
+++ b/hydragnn/utils/atomicdescriptors.py
@@ -1,0 +1,202 @@
+import os
+import torch
+import torch.nn.functional as F
+import json
+
+try:
+    import mendeleev
+except ImportError:
+    pass
+
+
+class atomicdescriptors:
+    def __init__(
+        self,
+        embeddingfilename,
+        overwritten=True,
+        element_types=["C", "H", "O", "N", "F", "S"],
+    ):
+        if os.path.exists(embeddingfilename) and not overwritten:
+            print("loading from existing file: ", embeddingfilename)
+            with open(embeddingfilename, "r") as f:
+                self.atom_embeddings = json.load(f)
+        else:
+            self.atom_embeddings = {}
+            if element_types is None:
+                self.element_types = []
+                for ele in mendeleev.get_all_elements():
+                    self.element_types.append(ele.symbol)
+            else:
+                self.element_types = []
+                for ele in mendeleev.get_all_elements():
+                    if ele.symbol in element_types:
+                        self.element_types.append(ele.symbol)
+                self.element_types = element_types
+            group_id = self.get_group_ids()
+            period = self.get_period()
+            covalent_radius = self.get_covalent_radius()
+            electron_affinity = self.get_electron_affinity()
+            block = self.get_block()
+            atomic_volume = self.get_atomic_volume()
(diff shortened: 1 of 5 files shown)


tests/test_atomicdescriptors.py
```diff
@@ -0,0 +1,26 @@
...
+
+import subprocess
+
+
+@pytest.mark.mpi_skip()
+def pytest_atomicdescriptors():
+    file_path = os.path.join(
+        os.path.dirname(__file__), "..", "hydragnn/utils/atomicdescriptors.py"
+    )
+    return_code = subprocess.call(["python", file_path])
```
allaffa, line 23: Very nice! I did not know about this.

hydragnn/utils/atomicdescriptors.py
```diff
@@ -0,0 +1,202 @@
...
+        group_id = F.one_hot(torch.Tensor(group_id).long(), num_classes=-1)
+        return group_id
+
+    def get_group_id(self, element):
+        if mendeleev.element(element).group_id is not None:
+            return mendeleev.element(element).group_id
+        else:  # 59-71
+            print("Warning: manually assign group_id for element: %s" % element)
+            if mendeleev.element(element).atomic_number < 71:
+                return mendeleev.element(element).atomic_number - 57 + 3
```
allaffa, line 82: Can you explain why you have to do this?

hydragnn/utils/atomicdescriptors.py
```diff
@@ -0,0 +1,202 @@
...
+        ie = F.one_hot(torch.Tensor(ie).long(), num_classes=10)
+        return ie
+
+    def get_atom_features(self, atomtype):
+        if isinstance(atomtype, str):
+            atomtype = mendeleev.element(atomtype).atomic_number
+        return torch.tensor(self.atom_embeddings[str(atomtype)])
+
+
+if __name__ == "__main__":
```
allaffa, line 198: Do we need this call? I think you used this to make sure that the code developed was working fine. But I don't think this main function should stay. Do you agree?

allaffa, line 198: I think the code would be better structures if this function were to be moved inside `test_atomicdescriptors.py`
Do you agree?

allaffa (commented):
First round of review

hydragnn/utils/atomicdescriptors.py
pzhanggit: `mendeleev` library returns `None` for those elements.

hydragnn/utils/atomicdescriptors.py
pzhanggit: Both would work. I prefer to have it here for convenient/easy testing/debugging. What do you think?

hydragnn/utils/atomicdescriptors.py
allaffa: Just throw an error if Mendeleev does not support information for a specific element, regardless of whether it is natural or synthetic.

allaffa (changes requested):

allaffa (approved):

PR #161, opened 2023-01-17
RawDatasetLoader turned into abstract class

allaffa:


allaffa:
@pzhanggit 

I followed your advice: I am assuming that the charge transfer has already been subtracted to the proton number before the data is imported. 
For now, you will see that the class `LSMS_RawDatasetLoader` still has the private method `__charge_density_update_for_LSMS`, but it is not used.

jychoi-hpc (commented):
It looks good to me. I have only a minor comment: would you like to clean up unused import statements in "lsms_raw_dataset_loader.py" and "cfg_raw_dataset_loader.py"?

allaffa:
> It looks good to me. I have only a minor comment: would you like to clean up unused import statements in "lsms_raw_dataset_loader.py" and "cfg_raw_dataset_loader.py"?

Yes. That's a good idea. Let me take care of that.

pzhanggit (commented):
It looks good to me. After the comment about charge density in lsms loader resolved, I am ready to approve the PR.

jychoi-hpc (approved):
Thank you for handling this. It looks good to me.

tests/deterministic_graph_data.py
```diff
@@ -130,6 +130,8 @@ def create_configuration(
         knn.fit(positions, node_feature)
         node_output_x = torch.Tensor(knn.predict(positions))
 
+    node_output_x = node_output_x + node_feature
+
     node_output_x_square = node_output_x ** 2
```
pzhanggit, line 135: Could you try something like below?
```suggestion
    node_output_x_square = node_output_x ** 2 + node_feature
```

allaffa: I already did. But that one also fails.

pzhanggit (approved):

PR #165, opened 2023-02-22
skip connection option added

allaffa:
As the name suggests, the skip connections in deep architecture bypass some of the neural network layers and feed the output of one layer as the input to the following levels. It is a standard module and provides an alternative path for the gradient with backpropagation.

Skip Connections were originally created to tackle various difficulties in various architectures and were introduced even before residual networks. In the case of residual networks or ResNets, skip connections were used to solve the degradation problems (e.g., vanishing gradient), and in the case of dense networks or DenseNets, it ensured feature reusability.

Since the residual block with the skip connection was shown to benefit the training of CNN, I thought it would be good providing it as optional also to stabilize the GNN (after all, the GNN architecture is a generalization of the CNN).

### DIFF
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -135,7 +135,7 @@ class Base(Module):
       
         self.layers = ModuleList()
         for i, (conv, batch_norm) in enumerate(zip(self.convs, self.batch_norms)):
-            if i > 0 and self.skip_connection:
+            if self.skip_connection:
                 self.layers.append(SkipConv(conv, batch_norm))
             else:
                 self.layers.append(Conv(conv, batch_norm))
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -18,7 +18,38 @@ from hydragnn.utils.model import loss_function_selection
 import sys
 from hydragnn.utils.distributed import get_device
 
+class Conv(torch.nn.Module):
+    """ conv module """
+    def __init__(self, conv, batch_norm, act=torch.nn.ReLU()):
+        super().__init__()
+        self.conv = conv
+        self.batch_norm = batch_norm
+        self.act = act
 
+    def forward(self, x=None, edge_index=None, edge_attr=None):
+        c = self.conv(x=x, edge_index=edge_index, edge_attr=edge_attr)
+        return self.act(self.batch_norm(c))
+
+class SkipConv(torch.nn.Module):
+    """ conv module with skip connection """
+    def __init__(self, conv, batch_norm, act=torch.nn.ReLU()):
+        super().__init__()
+        self.conv = conv
+        self.batch_norm = batch_norm
+        self.act = act
+
+    def forward(self, x=None, edge_index=None, edge_attr=None):
+        identity = x
+        c = self.conv(x=x, edge_index=edge_index, edge_attr=edge_attr)
+        return self.act(self.batch_norm(c)) + identity
+
+class ConvSequential(torch.nn.Sequential):
+    """ Extention to use graph inputs """
+    def forward(self, x=None, edge_index=None, edge_attr=None):
+        for module in self:
+            x = module(x=x, edge_index=edge_index, edge_attr=edge_attr)
+        return x
+    
 class Base(Module):
     def __init__(
         self,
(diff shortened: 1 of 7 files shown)


allaffa:
@pzhanggit @jychoi-hpc 
I put it as a WIP because the multi-heads of the `GATv2Conv` layer change the dimensional of the variable `x` across multiple message passing layers. If I set `self.heads=1`, everything works well also for this layer However, for any number of `self.heads` larger than 1, the code crashes. 
For now, the skip connection is allowed for all the layers except for `GATv2Conv`. I would like to probe more into the  `GATv2Conv` layers and see if there is a practical way to apply the skip connection also for this layer.

allaffa:
@pzhanggit @jychoi-hpc 
I activated this PR to kick off the review process :)

jychoi-hpc (approved):
It looks good to me. As long as we solve the conflict, I think it is ok to merge.

hydragnn/models/Base.py
```diff
@@ -100,6 +140,15 @@ def __init__(
         if self.initial_bias is not None:
             self._set_bias()
 
+        self.layers = ModuleList()
+        for i, (conv, batch_norm) in enumerate(zip(self.convs, self.batch_norms)):
```
pzhanggit, line 144: Remove `i` and `enumerate` since not used anymore

hydragnn/models/Base.py
```diff
@@ -244,6 +293,8 @@ def _multihead(self):
     def forward(self, data):
         x = data.x
 
+        count_conv_layers = 0
+
```
pzhanggit, line 297: Not needed?

allaffa:
> It looks good to me. As long as we solve the conflict, I think it is ok to merge.

@jychoi-hpc I manually solved the conflicts and rebase to maintain a linear history on the main branch. Would you mind re-controlling that I di don't remove anything of what you did? Thanks.

hydragnn/models/Base.py
```diff
@@ -19,6 +19,44 @@
 from hydragnn.utils.distributed import get_device
 
 
+class Conv(torch.nn.Module):
```
pzhanggit, line 22: Do we want to move the three new classes to the end of the file?

hydragnn/models/create.py
```diff
@@ -182,6 +188,7 @@ def create_model(
             freeze_conv=freeze_conv,
             num_conv_layers=num_conv_layers,
             num_nodes=num_nodes,
+            skip_connection=skip_connection,
         )
 
     elif model_type == "SchNet":
```
pzhanggit, line 194: `skip_connection` also for `SchNet`?

pzhanggit (commented):
Look good to me too. I only have a few minor comments. Will approve once they are fixed and the tests pass.

hydragnn/models/Base.py
allaffa: The new classes are important building blocks used in the rest of the script. 
Logically, I like more when the building blocks are defined first, before they are used by any other function or class. 
Does it make sense?

pzhanggit: > The new classes are important building blocks used in the rest of the script. Logically, I like more when the building blocks are defined first, before they are used by any other function or class. Does it make sense?

Sounds good to me too.

hydragnn/models/Base.py
```diff
@@ -100,6 +140,15 @@ def __init__(
...
             self._set_bias()
 
+        self.layers = ModuleList()
+        for conv, batch_norm in zip(self.convs, self.batch_norms):
+            if self.skip_connection:
+                self.layers.append(SkipConv(conv, batch_norm))
+            else:
+                self.layers.append(Conv(conv, batch_norm))
+
+        self.conv_shared = ConvSequential(*self.layers)
```
pzhanggit, line 150: Hmmmm,  just noticed that `self.conv_shared` is defined but never used in `forward` function? Still the old one `self.convs` is used.

pzhanggit (changes requested):
Need to update `forward` function. That skip connection is defined in `self.conv_shared` but not called in `forward` function. It is still calling the old `self.convs`.

hydragnn/models/Base.py
allaffa: This is a mistake of mine. In merging manually I definitely messed up something. 
@jychoi-hpc would you mind checking ? I had to` force push`. Therefore, locally what you did does not show up anymore.

jychoi-hpc: I pushed the chagnes for the forward function to your local repo: https://github.com/allaffa/HydraGNN/pull/3 Once you merge, it will show up here.

allaffa:
@pzhanggit @jychoi-hpc 
Thanks. I turned on again the PR for an additional review :-)

PR #166, opened 2023-02-23
Stop training if val_loss is not decreasing

pzhanggit:


hydragnn/train/train_validate_test.py
```diff
@@ -148,6 +159,14 @@ def train_validate_test(
                 output_names=config["Variables_of_interest"]["output_names"],
                 iepoch=epoch,
             )
+        if EarlyStop:
+            if earlystopper(reduce_values_ranks(val_loss)):
```
allaffa, line 163: @pzhanggit 
I know that the comment I am going to write has nothing to do with this PR. In fact, I am going to approve this PR right away. 
However `reduce_values_rank` right now is a hard-coded implementation of averaging across processes. We may consider in the future making it more flexible to handle arbitrarily difference reduce operations (max, min, sum, average)

allaffa (approved):

PR #175, opened 2023-05-05
Smiles utils includes atomic descriptor object as optional argument

allaffa:


allaffa:
@pzhanggit @jychoi-hpc let me know what you think

jychoi-hpc (approved):
It looks good to me.

PR #178, opened 2023-06-12
DimeNet Model

JustinBakerMath:
Introduces the key components of the [DimeNet](https://arxiv.org/abs/2003.03123) model to the HydraGNN library by modifying the [PyG implementation](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.models.DimeNet.html#torch_geometric.nn.models.DimeNet).

The core addition is the [DIMEStack](hydragnn/models/DIMEStack.py) which is inherits from `Base` to provide convolutional layers on request using the `_get_conv()` method.

*Notice:* Because the PyG implementations use the [glorot](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/inits.py#LL30C9-L30C9) initialization, the construction of the hidden dimension must be larger than 1 to avoid division by zero. Only the `input_dim` and `output_dim` are provided to `_get_conv()`, without constraint on the dimension of the convolution. Therefore, DIMEStack will use the `input_dim` if larger than 1 or the `output_dim` otherwise. If neither is larger than one, then the [assertion](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-87e602377a1c636d90174b0b40ce5e1d278c535cde35936c28c164df27cf1dc9R69) will fail.

The DIMEStack also computes the necessary convolutional arguments, being
    `rbf`: (tensor) the radial basis function output.
    `sbf`: (tensor) the spherical basis function output.
    `i`: The message passing target.
    `j`: The message passing source.
    `idx_kj`: A portion of the bond triple.
    `idx_ji`:  The second component of the bond triple.

**Notice:** DimeNet uses the `triplet()` method to compute the bond angles between triplets of nodes. This information is computed on every call but there is room for computational speed ups by computing once and then updating iteratively.

The stack modifies the `Embedding` and `InteractionBlock` and preserves the`OutputBlock` of the baseline DimeNet implementation. For the `Embedding` layer, the baseline atomic feature embedding is [removed](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-87e602377a1c636d90174b0b40ce5e1d278c535cde35936c28c164df27cf1dc9R166) as embedding is handled by HydraGNN.

The `InteractionBlock` layer does not find the optimal `einsum` path when using the baseline `einsum` so it is [segmented](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-87e602377a1c636d90174b0b40ce5e1d278c535cde35936c28c164df27cf1dc9R190-R191) into two equivalent steps.

Six hyperparameters are [introduced](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-d0727877396ee72bc17233de8866f3cf9fef34b1cdbc654da48e7dd66965f94cR51-R56).

- `num_before_skip`/`num_after_skip` control layers before and after the attention mechanism in the `OutputBlock`.
- `num_bilinear` determines the number of bilinear layers used in the `InteractionBlock`
- `num_radial`/`envelope_exponent` determine the radial basis functions used.
- `num_spherical` determines the number of spherical basis functions used

These arguments are [checked](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-fbd92e9a9b3c8780b461a137b36ce542a7f02b23737901b96d3dc47f84ecbd9fR60-R71) to exist or set to none, and [verified](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-d0727877396ee72bc17233de8866f3cf9fef34b1cdbc654da48e7dd66965f94cR223-R231) not to be none before use by DimeNet.

Other updates include documentation and testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ There are many options for HydraGNN; the dataset and model type are particularly
 important:
  - `["Verbosity"]["level"]`: `0`, `1`, `2`, `3`, `4`
  - `["Dataset"]["name"]`: `CuAu_32atoms`, `FePt_32atoms`, `FeSi_1024atoms`
- - `["NeuralNetwork"]["Architecture"]["model_type"]`: `PNA`, `MFC`, `GIN`, `GAT`, `CGCNN`, `SchNet`
+ - `["NeuralNetwork"]["Architecture"]["model_type"]`: `PNA`, `MFC`, `GIN`, `GAT`, `CGCNN`, `SchNet`, `DimeNet`
 
 ### Citations
 "HydraGNN: Distributed PyTorch implementation of multi-headed graph convolutional neural networks", Copyright ID#: 81929619
--- a/hydragnn/models/DIMEStack.py
+++ b/hydragnn/models/DIMEStack.py
@@ -1,0 +1,202 @@
+##############################################################################
+# Copyright (c) 2021, Oak Ridge National Laboratory                          #
+# All rights reserved.                                                       #
+#                                                                            #
+# This file is part of HydraGNN and is distributed under a BSD 3-clause      #
+# license. For the licensing terms see the LICENSE file in the top-level     #
+# directory.                                                                 #
+#                                                                            #
+# SPDX-License-Identifier: BSD-3-Clause                                      #
+##############################################################################
+
+from typing import Callable, Optional, Tuple
+from torch_geometric.typing import SparseTensor
+
+import torch
+from torch import Tensor
+from torch.nn import SiLU
+
+from torch_geometric.nn import Linear, Sequential
+from torch_geometric.nn.models.dimenet import (
+    BesselBasisLayer,
+    EmbeddingBlock,
+    InteractionBlock,
+    OutputBlock,
+    SphericalBasisLayer,
+)
+from torch_geometric.utils import scatter
+
+from .Base import Base
+
+
+class DIMEStack(Base):
+    """
+    Generates angles, distances, to/from indices, radial basis
+    functions and spherical basis functions for learning.
+    """
+
+    def __init__(
+        self,
+        envelope_exponent,
+        num_after_skip,
+        num_before_skip,
+        num_bilinear,
+        num_radial,
+        num_spherical,
+        radius,
+        *args,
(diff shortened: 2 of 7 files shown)


JustinBakerMath:
- First, with regard to formatting. I would like to follow the [pre-commit instructions](https://github.com/ORNL/HydraGNN/blob/main/CONTRIBUTING.md), but I encounter known [compatibility issues](https://github.com/psf/black/issues/2964) between `click` and `black`. If an old version of `black` is being used, is it possible to get a `click` version [pinned](https://github.com/psf/black/issues/2964#issuecomment-1080974737) in the [pre-commit configuration](https://github.com/ORNL/HydraGNN/blob/main/.pre-commit-config.yaml).
- Second, the current errors [1](https://github.com/ORNL/HydraGNN/actions/runs/5247683121/jobs/9478161664#step:8:134) and [2](https://github.com/ORNL/HydraGNN/actions/runs/5247683121/jobs/9478161664#step:8:162) indicate import issues inside of `torch_geometric` with the `sympy` dependency. Trouble shooting this error is harder, as I am relying on the `torch_geometric` [install](https://github.com/ORNL/HydraGNN/blob/main/requirements-torchdep.txt#L1) in the [CI](https://github.com/ORNL/HydraGNN/blob/main/.github/workflows/CI.yml#L43) to build the dependencies. The most brute-force option is to re-write the [SphericalBasisLayer](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/models/dimenet.py#L82) to be independent of `sympy`.  The opposite end seems to be to include `sympy` in one of the `requirements-*.txt`.
- I've included a screenshot of my local pytest output.

DimeNetPassing

JustinBakerMath:
Closed in favor of DimeNet++ integration and request to include test results for non-unit test datasets.

PR #180, opened 2023-06-14
gather_deg_dist needs to map 'deg' variable to GPUs before calling NCCL gathering operations

allaffa:
All the data loaders use the `pin_memory=True`, which can work only if the data is stored on CPUs. 
If the entire data were stored on the GPU, `pin_memory=True` would make the code crash

Since all our data is stored on CPU, the `max_de`g and `deg ` variables are also stored on the CPU. Therefore, the NCCL gathering operations crash because they require the variables to be on the device. 

This PR fixed this problem.


allaffa:
@pzhanggit @jychoi-hpc this PR is really for review.

jychoi-hpc (approved):
I haven't tested on Summit but It looks good to me for now to merge. I will do more through test later.

PR #181, opened 2023-06-17
Init examples on cpu for memory pinning

JustinBakerMath:
Keep examples on cpu in the pre_transform, so that they properly interact with memory pinning when being migrated to the gpu.


allaffa (approved):

allaffa:
@jychoi-hpc 
Justin and I had to make this small modification to the QM9 and MD17 examples because the class inheritance of the dataset classes we have implemented in HydraGNN require to always use memory pinning, which assumes the data to be stored on the host. 
Since these examples were locating directly the data on the device, we had to disable this option and force the data to reside on the host.

PR #182, opened 2023-06-19
Minimum Requirement for Scatter

JustinBakerMath:


allaffa (approved):

allaffa (commented):
@jychoi-hpc 
would increasing the version of PyTorch required break any wheel building to run HydraGNN on OLCF supercomputers?

allaffa (commented):
I need to make sure that this dependency requirement does not cause problems on the supercomputers, where unfortunately the software stack is not alway the most up-to-date

JustinBakerMath:
> I need to make sure that this dependency requirement does not cause problems on the supercomputers, where unfortunately the software stack is not alway the most up-to-date

I completely understand. Let me know if going forward it makes sense for me to add all three of you as reviewers whenever I make version related PRs.

The brute force alternative to this PR is to hard code `scatter` when I need it for DimeNet.

jychoi-hpc (approved):
We are using 1.13. It looks good to me.

PR #184, opened 2023-06-22
Implementation of DimeNet++

JustinBakerMath:
Introduces the key components of the [DimeNet++](https://arxiv.org/abs/2011.14115) model to the HydraGNN library by modifying the [PyG implementation](https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.nn.models.DimeNetPlusPlus.html#torch_geometric.nn.models.DimeNetPlusPlus).

The core addition is the [DIMEStack](hydragnn/models/DIMEStack.py) which is inherits from `Base` to provide convolutional layers on request using the `_get_conv()` method.

*Notice:* Because the PyG implementations use the [glorot](https://github.com/pyg-team/pytorch_geometric/blob/master/torch_geometric/nn/inits.py#LL30C9-L30C9) initialization, the construction of the hidden dimension must be larger than 1 to avoid division by zero. Only the `input_dim` and `output_dim` are provided to `_get_conv()`, without constraint on the dimension of the convolution. Therefore, DIMEStack will use the `input_dim` if larger than 1 or the `output_dim` otherwise. If neither is larger than one, then the [assertion](https://github.com/ORNL/HydraGNN/compare/main...JustinBakerMath:HydraGNN:model-dimenet?expand=1#diff-87e602377a1c636d90174b0b40ce5e1d278c535cde35936c28c164df27cf1dc9R69) will fail.

The DIMEStack also *computes* the necessary convolutional arguments, being
    `rbf`: (tensor) the radial basis function output.
    `sbf`: (tensor) the spherical basis function output.
    `i`: The message passing target.
    `j`: The message passing source.
    `idx_kj`: A portion of the bond triple.
    `idx_ji`:  The second component of the bond triple.

**Notice:** DimeNet uses the `triplet()` method to compute the bond angles between triplets of nodes. This information is computed on every call but there is room for computational speed ups by computing once and then updating iteratively.

The stack modifies the `Embedding` , the baseline atomic feature embedding is removed and embedding is handled by HydraGNN.

Eight hyperparameters are introduced.

- `basis_emb_size`/`int_emb_size`/`out_emb_size` determines the dimensions of the embedding mechanism.
- `num_before_skip`/`num_after_skip` control layers before and after the attention mechanism in the `OutputBlock`.
- `num_radial`/`envelope_exponent` determine the radial basis functions used.
- `num_spherical` determines the number of spherical basis functions used

Other updates include documentation and testing.


requirements.txt
```diff
@@ -6,3 +6,4 @@ ase
 tqdm
 tensorboard
 psutil
+sympy
```
allaffa, line 9: @jychoi-hpc 
Do you think this requirement on `sympy` may represent a problem on OLCF machines? 
Do you suggest that we move this additional dependency in a file that makes the `sympy` requirement not mandatory?

jychoi-hpc: It should not be a problem. sympy is pretty a lightweight.

jychoi-hpc (approved):

allaffa (approved):

PR #187, opened 2023-06-26
Checkpoint Training Model

JustinBakerMath:
Basic checkpointing for training based on the synchronized `validation_loss` performance metric.

Can be turned on by adding
```
"Training" : {
    "Checkpoint" : true,
}
```
with additional `checkpoint_freq` argument for the frequency of checkpointing (`default=10`)
```
"Training" : {
    "Checkpoint" : true,
    "checkpoint_freq" : 10,
}
```

### DIFF
--- a/hydragnn/utils/model.py
+++ b/hydragnn/utils/model.py
@@ -165,22 +165,24 @@ class Checkpoint:
     """
     Checkpoints the model and optimizer when:
         + The count exceeds the checkpointing frequency
-        + [Optional] The performance metric is smaller than the stored performance metric
+        + The performance metric is smaller than the stored performance metric
     Args
       frequency: (int) Frequency of checkpointing.
       path: (str) Path for checkpointing
       name: (str) Model name for the directory and the file to save.
-      diable_perf_metric: (bool) optionally disable the performance metric check.
     """
 
     def __init__(
-        self, frequency=10, path="./logs/", name="model.pk", disable_perf_metric=False
+        self,
+        frequency: int = 10,
+        path: str = "./logs/",
+        name: str = "model.pk",
     ):
         self.count = 0
         self.frequency = frequency
         self.path = path
         self.name = name
-        self.min_perf_metric = float("-inf") if disable_perf_metric else float("inf")
+        self.min_perf_metric = float("inf")
         self.min_delta = 0
 
     def __call__(self, model, optimizer, perf_metric):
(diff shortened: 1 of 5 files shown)


tests/inputs/ci.json
```diff
@@ -68,6 +68,8 @@
             "perc_train": 0.7,
 	    "EarlyStopping": true,
 	    "patience": 10,
+        "Checkpoint": true,
```
allaffa, line 71: In terms of indentation, it is best that checkpoint be encapsulated within "Training"

tests/inputs/ci.json
```diff
@@ -66,8 +66,10 @@
         "Training": {
             "num_epoch": 100,
             "perc_train": 0.7,
-	    "EarlyStopping": true,
-	    "patience": 10,
+            "EarlyStopping": true,
```
allaffa, line 69: Would you mind adding this section to all the CI.json file where at least 10 epochs are performed?

allaffa (changes requested):
We should add the additional input to all the JSON file of CI tests that perform at least 10 epochs.

JustinBakerMath:
After revisiting this: `checkpoint_freq` and `min_perf_metric` aren't ideally compatible.

If the goal is to use `min_perf_metric` to checkpoint, then there should be a *warmup* rather than a frequency.

I've refactored this, replacing `checkpoint_freq` with `checkpoint_warmup` for a *warmup* to checkpointing.

Two print lines are also included
- (every epoch) indicate best performance metric 
- (upon saving) indicate new checkpoint with performance metric.

hydragnn/utils/model.py
```diff
@@ -159,3 +159,39 @@ def __call__(self, val_loss):
...
+        self.count = 1
+        self.warmup = warmup
+        self.path = path
+        self.name = name
+        self.min_perf_metric = float("inf")
+        self.min_delta = 0
+
+    def __call__(self, model, optimizer, perf_metric):
+
+        if (perf_metric > self.min_perf_metric + self.min_delta) or (
```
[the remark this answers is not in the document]
allaffa: The call to `save_model` has no input parameter that varies over the iteration count. How do you make sure that different checkpoints are saved in directories with different path names?

allaffa:
@jychoi-hpc 

Hi Jong. Justin implemented a checkpoint restart which saves a copy of the model when the validation loss function gets reduced at a new iteration. A warm-up waiting is applied at earlier iterations of the training to avoid saving too many models at the beginning. 
I was wondering if the PR may cause too many bottlenecks with significant slow-down for training. 
Any thoughts on this?

jychoi-hpc (approved):
I checked on Summit. It works for me. Since users can control (delay, etc) through json file, I think it is ok.

PR #188, opened 2023-07-11
call to data transforms fixed

allaffa:


allaffa:
@jychoi-hpc @pzhanggit 
May you take a look at this PR?

pzhanggit (approved):
It looks good to me. Two minor things:
1. not sure if intended is that edge length will be one of the edge attributes together with the others in `PointPairFeatures()`. And if we do not want this for both `Spherical` and `PointPairFeatures`, we can set `cat=False` when creating/initializing the objects (then change in line 362 is not needed).
2. maybe merge `spherical_coordinates_transform` and `point_pair_features_transform` into one, like `edge_features_transform`? then lines 386-389 can be merged into one line.

allaffa:
> It looks good to me. Two minor things:
> 
> 1. not sure if intended is that edge length will be one of the edge attributes together with the others in `PointPairFeatures()`. And if we do not want this for both `Spherical` and `PointPairFeatures`, we can set `cat=False` when creating/initializing the objects (then change in line 362 is not needed).
> 2. maybe merge `spherical_coordinates_transform` and `point_pair_features_transform` into one, like `edge_features_transform`? then lines 386-389 can be merged into one line.

@pzhanggit 
Thank you! That actually was a mistake of mine. 
Both spherical coordinates and point pair features are vector edge features, and each one of them includes the edge length as a scalar feature. 
A naive simultaneous utilization of both spherical coordinates and point point features together includes the distance multiple times in the edge-feature vector. 
For now, I force the code to crash if they are both used.

hydragnn/utils/abstractrawdataset.py
```diff
@@ -120,6 +120,12 @@ def __init__(self, config, dist=False, sampling=None):
                     "PointPairFeatures"
                 ]
 
+        if self.spherical_coordinates:
+            self.spherical_coordinates_transform = Spherical(norm=False)
+
+        if self.point_pair_features:
+            self.point_pair_features_transform = PointPairFeatures()
```
pzhanggit, line 127: Wondering if something like
```suggestion
        if self.spherical_coordinates:
            self.edge_feature_transform = Spherical(norm=False, cat=False)

        if self.point_pair_features:
            self.edge_feature_transform = PointPairFeatures(cat=False)
```
is simpler, then check if `hasattr()` in line 263, and update/merge lines 389-403.

pzhanggit (commented):
Thanks, the new changes work for me and potentially can be simplified tho. Let me know what you think.

jychoi-hpc (approved):
It looks good to me.

pzhanggit (approved):

PR #189, opened 2023-08-15
Use Adios BP4

jychoi-hpc:
Adios recently added new BP5 format. But, it is not compatible with what we are using in HydraGNN. 
We will use BP4 until we figure out the correct way to use BP5.

Another fix is about `sys.setrecursionlimit(1000000)`. It was needed when I create a large ising dataset.

### DIFF
--- a/examples/ising_model/train_ising.py
+++ b/examples/ising_model/train_ising.py
@@ -223,6 +223,7 @@ if __name__ == "__main__":
         3. Split into a train, valid, and test set (split_dataset)
         4. Save as Adios file in parallel
         """
+        sys.setrecursionlimit(1000000)
         dir = os.path.join(os.path.dirname(__file__), "./dataset/%s" % modelname)
         if rank == 0:
             if os.path.exists(dir):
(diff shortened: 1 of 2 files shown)


kshitij-v-mehta:
Could you let me know why BP5 is not compatible with HydraGNN?

jychoi-hpc:
Current python code is not working with BP5. It doesn't read values correctly. It looks like the grammar is changed.
For example, we use the following code:
```
with ad2.open(self.filename, "r", MPI.COMM_SELF) as f:
    self.vars = f.available_variables()
    self.ndata = f.read_attribute("%s/ndata" % label).item()
```
But, I don't get the correct values with BP5.  I have to do some extra things but it doesn't look right.

@kshitij-v-mehta  do you have any python example code for using BP5?

kshitij-v-mehta:
Indeed the API in BP5 has changed. You have two options: use `rra` instead of `r` in the `ad2.open` call, or use the low-level API that has BeginStep and EndStep before you inquire variables. Norbert suggests it is better to use BeginStep and EndStep.

jychoi-hpc:
I found a solution to use both "BP4" and "BP5".

pzhanggit (approved):
Looks good to me.

allaffa (approved):

PR #190, opened 2023-08-23
utilies to update data objects moved to utils

allaffa:


jychoi-hpc (approved):
It looks good to me.

pzhanggit (commented):
The changes look good to me. 
Could you do a grep search to make sure if all calls related to `update_predicted_values` and `update_atom_features` are updated accordingly? Thanks.

allaffa:
> The changes look good to me.
> Could you do a grep search to make sure if all calls related to `update_predicted_values` and `update_atom_features` are updated accordingly? Thanks.

@pzhanggit 
Thanks. Indeed, I had forgotten to change it inside `abstractrawdataset.py`. I noticed that the class inheritance from `abstractrawdataset.py` was not guarded. Therefore, I created a `pytest ` to check this. Th new pytest is called `test_datasetclass_inheritance.py`

allaffa:
@pzhanggit 

I followed your suggestions about running ‘grep’, and indeed  I had forgotten to change it inside abstractrawdataset.py. I noticed that the class inheritance from abstractrawdataset.py was not guarded.
 
Therefore, I created a pytest to check that abstractrawdataset.py is used properly.
The new pytest is called test_datasetclass_inheritance.py
 
The test runs just fine on my laptop. However, it fails on the CI test with the following error:
AttributeError: 'LSMSDataset' object has no attribute 'edge_feature_transform'
 
When I run:
 
grep -n -r ‘edge_feature_transform’ ./
 
the command is empty.
 
Given that I edge_feature_transform is never called anywhere in the code, I do not understand why this attributed, which is never invoked, could cause errors.
 
Would you mind helping me out understand what is going on?

jychoi-hpc:
I ran on my local machine and found test_datasetclass_inheritance.py is causing a problem. It looks like there is a conflict in using ddp inside pytest with mutliple test cases. I will look for some examples in the internet.

allaffa:
@jychoi-hpc @pzhanggit 
Wait for the CI test to end, and then you can proceed reviewing this PR one more time :-)

pzhanggit (approved):

PR #191, opened 2023-09-09
pickle formatting fixed for OGB dataset

allaffa:
The OGB examples was using obsolete functionalities. 
I updated the pre-processing for `pickle` files. 
@jychoi-hpc would you mind updating the functionalities for ADIOS to make them match with the new formatting requested?


jychoi-hpc (approved):
It is good. I tested on my machine and it works.

PR #192, opened 2023-09-16
restructure pre-process data utils

allaffa:
The function `update_atom_features` is already available in the file `hydragnn\preprocess\utils.py`.
Therefore, there is no need to have a re-implementation as method inside the class `abstractrawdataset`.

The function `stratified_sampling` is a general functionality, and its use it not restricted to only datasets created with the class `abstractrawdataset`.

### DIFF
--- a/hydragnn/preprocess/utils.py
+++ b/hydragnn/preprocess/utils.py
@@ -290,3 +290,48 @@ def update_atom_features(atom_features: [AtomFeatures], data: Data):
     """
     feature_indices = [i for i in atom_features]
     data.x = data.x[:, feature_indices]
+
+
+def stratified_sampling(dataset: [Data], subsample_percentage: float, verbosity=0):
+    """Given the dataset and the percentage of data you want to extract from it, method will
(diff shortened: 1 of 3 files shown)


hydragnn/preprocess/utils.py
```diff
@@ -290,3 +290,47 @@ def update_atom_features(atom_features: [AtomFeatures], data: Data):
     """
     feature_indices = [i for i in atom_features]
     data.x = data.x[:, feature_indices]
+
+
+def stratified_sampling(dataset: [Data], subsample_percentage: float, verbosity=0):
```
pzhanggit, line 295: I am not sure if we want to solve it in this PR, but thinking we should make the index/labels used in the stratified split flexible, maybe provided by users.

pzhanggit (commented):
Looks good to me. Left one comment on labels used in stratified splitting.

hydragnn/preprocess/utils.py
pzhanggit: Yeah, sounds good to me.

pzhanggit (approved):

PR #194, opened 2023-10-23
Incorporating Equivariance

JustinBakerMath:
A reformatting of the Base class and the inherited classes to support equivariance through updating positional data (`data.pos`) in each graph convolution.


hydragnn/models/EGCLStack.py
```diff
@@ -184,7 +206,7 @@ def coord_model(self, coord, edge_index, coord_diff, edge_feat):
             trans, min=-100, max=100
         )  # This is never activated but just in case it case it explosed it may save the train
         agg = unsorted_segment_mean(trans, row, num_segments=coord.size(0))
-        coord += agg * self.coords_weight
+        coord = coord + agg * self.coords_weight
```
allaffa, line 209: isn't this equivalent to 

`coord += agg * self.coords_weight`

which is what was already there before?

allaffa (commented):
Just have one little questions.

hydragnn/models/EGCLStack.py
JustinBakerMath: Yes, this is identical. However, PyTorch discourages in-place operations as noted [here.](https://pytorch.org/tutorials/beginner/blitz/tensor_tutorial.html#operations)

> In-place operations save some memory, but can be problematic when computing derivatives because of an immediate loss of history. Hence, their use is discouraged.

The updated version is improved in this regard.

allaffa (approved):

hydragnn/models/SCFStack.py
```diff
@@ -40,48 +44,188 @@ def __init__(
...
+        agg = unsorted_segment_mean(trans, row, num_segments=coord.size(0))
+        coord = coord + agg
+        return coord
+
+    def reset_parameters(self):
+        torch.nn.init.xavier_uniform_(self.lin1.weight)
+        torch.nn.init.xavier_uniform_(self.lin2.weight)
+        self.lin2.bias.data.fill_(0)
+
+    def forward(
```
pzhanggit, line 186: It seems that when `equivariant`, the `pos` is not affecting the node feature embeddings, `x`, so how could this changes the model? Am I missing something here?

allaffa: `equivariant` should update `x` and `pos` independently, without necessarily having one affect the update of the other. However, I will let @JustinBakerMath give you a more thorough explanation.

pzhanggit: Thank you, Max. Then how could the model improve the final predictions that will be based on some global pooling values of `x`?

allaffa: The pooling values of `x` happens only for quantities that are referred to the total graph. 
Node-level properties do not require the global pooling. 

Total-graph properties should be invariant. If I rotate or translate an atomic structure, the total-graph property should be the same. 

Nodal properties should be equivariant. If I rotate or translate an atomic structure, the node-level properties should rotate or translate by the same degree.

pzhanggit: I agree with all you said, but I am not sure they address question. My question is: it seems with the current implementation, it will not change the prediction results with `equivariant` being turned on or off. In particular, `pos` is updated when `equivariant` on, but it is not used in updating nodal embedding features (and hence not changing final predictions, unless I miss something)?  

The implementation in `EGCLStack` (https://github.com/ORNL/HydraGNN/blob/ccbdddc14ebbe7412ba3304b929d9568c13639aa/hydragnn/models/EGCLStack.py#L223) makes more sense to me, where `coord` is used to update `x` and hence will impact the final prediction results.

Happy to explain more if I am not clear and learn what I miss here.

allaffa: Now I see what you mean. The SchNet model (as it was proposed in the literature) uses only distances (edge lengths) to create the graph embedding. I guess this is why the SCFStack (which is used to build a SchNet model) does not explicitly use the coordinates `pos` to update `x` at each message passing layer. 
@JustinBakerMath may you confirm? maybe we should put a comment line to make this clear?

hydragnn/models/SCFStack.py
```diff
@@ -10,9 +10,13 @@
 ##############################################################################
 
 from typing import Optional
+from math import pi as PI
 
 import torch
-from torch.nn import Identity, Linear, Sequential
+from torch import Tensor
+from torch.nn import Identity, Linear, ReLU, Sequential
+from torch_geometric.nn import Sequential as PyGSeq
```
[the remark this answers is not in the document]
JustinBakerMath: The two methods have different functionalities.

hydragnn/models/SCFStack.py
```diff
@@ -40,48 +44,188 @@ def __init__(
...
+        coord_diff = coord[row] - coord[col]
+        radial = torch.sum((coord_diff) ** 2, 1).unsqueeze(1)
+
+        norm = torch.sqrt(radial) + 1
+        coord_diff = coord_diff / (norm)
+
+        return radial, coord_diff
+
+
+def unsorted_segment_mean(data, segment_ids, num_segments):
```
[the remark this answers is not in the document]
JustinBakerMath: There currently aren't any clear utility files to put common methods in, but perhaps we can add a `utils.py` file in `hydragnn/models/` @allaffa .

hydragnn/models/SCFStack.py
JustinBakerMath: The position is used to update `x` but through the generation of the relative distances 
https://github.com/ORNL/HydraGNN/blob/ccbdddc14ebbe7412ba3304b929d9568c13639aa/hydragnn/models/SCFStack.py#L99
which are used for the Gaussian embedding
https://github.com/ORNL/HydraGNN/blob/ccbdddc14ebbe7412ba3304b929d9568c13639aa/hydragnn/models/SCFStack.py#L100
which are used in the embedding block
https://github.com/ORNL/HydraGNN/blob/ccbdddc14ebbe7412ba3304b929d9568c13639aa/hydragnn/models/SCFStack.py#L101

hydragnn/models/SAGEStack.py
```diff
@@ -23,10 +23,25 @@ def __init__(self, *args, **kwargs):
...
             in_channels=input_dim,
             out_channels=output_dim,
         )
 
+        input_args = "x, pos, edge_index"
+        conv_args = "x, edge_index"
+
+        if self.use_edge_attr:
+            input_args += ", edge_attr"
+            conv_args += ", edge_attr"
```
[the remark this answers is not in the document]
JustinBakerMath: Thank you for pointing these out, I've removed them here: https://github.com/ORNL/HydraGNN/commit/7331c8dd02ecd908fd612f61197cb6c367d96a7c

hydragnn/models/SCFStack.py
pzhanggit: Thanks! I missed that part!

allaffa:
@JustinBakerMath we need to resolve some conflicts with the new version of the branch since I merged PR#195 (activation functions). Let me know if you want me to take care of the rebasing myself. Hopefully, this will be pretty straightforward to take care of.

hydragnn/models/SCFStack.py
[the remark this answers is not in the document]
allaffa: @JustinBakerMath 
We have a directory called `hydragnn/utils` that is already dedicated to all the possible utilities. 
Inside of this directory, we have a file called `models.py` that collects all the utilities that support the definition of a model. Why don't we use that file?

allaffa:
> I was able to run a few examples without errors. But, I am not sure if I really tested this new equivariance feature. I am just wondering if there is any example in the list of our examples to check this new feature. I am thinking simply turning on "equivariance: true" and changing "model_type: EGNN or SchNet".
> 
> Otherwise, it looks good. I ran unit tests (no mpi) and they went ok in my local machine.

@jychoi-hpc  
yes, we can use the `equivariant` for the QM9 dataset.

JustinBakerMath:
> @JustinBakerMath we need to resolve some conflicts with the new version of the branch since I merged PR#195 (activation functions). Let me know if you want me to take care of the rebasing myself. Hopefully, this will be pretty straightforward to take care of.

I have now merged the changes to the activation functions to the changes for equivariance.

allaffa:
> > @JustinBakerMath we need to resolve some conflicts with the new version of the branch since I merged PR#195 (activation functions). Let me know if you want me to take care of the rebasing myself. Hopefully, this will be pretty straightforward to take care of.
> 
> I have now merged the changes to the activation functions to the changes for equivariance.

@JustinBakerMath Thanks. I will run the `equivariance` with the QM9 dataset. @jychoi-hpc would you mind doing the same?
The best example for this would be the `GDB-9-Ex` dataset. However, the GDB-9-Ex dataset is not updated in the current version of he main branch. So I think that using the standard QM9 is the best thing to do.

pzhanggit (approved):
Looks good to me, thanks!

allaffa:
> Looks good to me, thanks!

@pzhanggit, thanks! I just asked @JustinBakerMath to move `unsorted_segment_mean` into `hydragnn/utils/models.py`.

allaffa:
> > I was able to run a few examples without errors. But, I am not sure if I really tested this new equivariance feature. I am just wondering if there is any example in the list of our examples to check this new feature. I am thinking simply turning on "equivariance: true" and changing "model_type: EGNN or SchNet".
> > Otherwise, it looks good. I ran unit tests (no mpi) and they went ok in my local machine.
> 
> @jychoi-hpc yes, we can use the `equivariant` for the QM9 dataset.

@jychoi-hpc I ran the QM9 example in debug mode using PyCharm with `model_type = "EGNN"` and `equivariance = True`. I could see the equivariant message passing being activated.

jychoi-hpc:
Great. I was able to run too with `model_type = "EGNN"` and `equivariance = True` for QM9. This PR looks good to me.

PR #202, opened 2023-11-05
abstractrawdataset updated

allaffa:
This PR addresses two issues:

- `self.__normalize_dataset()` is safeguarded by a boolean variable and used only when the user specifies it in the .JSON file
- fixes a bug that was hidden inside the `AbstractRawDataset` .
`update_predicted_values` and `update_atom_features` are already included in the SimplePickleDataset class. Therefore, there is no need to used them now inside `AbstractRawDataset`

### DIFF
--- a/hydragnn/utils/abstractrawdataset.py
+++ b/hydragnn/utils/abstractrawdataset.py
@@ -62,6 +62,11 @@ class AbstractRawDataset(AbstractBaseDataset, ABC):
         """
 
         # self.serial_data_name_list = []
+        self.normalize_features = (
+            config["Dataset"]["normalize_features"]
+            if config["Dataset"]["normalize_features"] is not None
+            else False
+        )
         self.node_feature_name = (
             config["Dataset"]["node_features"]["name"]
             if config["Dataset"]["node_features"]["name"] is not None
@@ -205,7 +210,8 @@ class AbstractRawDataset(AbstractBaseDataset, ABC):
         # scaled features by number of nodes
         self.__scale_features_by_num_nodes()
 
-        self.__normalize_dataset()
+        if self.normalize_features:
+            self.__normalize_dataset()
 
     def __normalize_dataset(self):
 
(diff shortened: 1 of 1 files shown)


allaffa:
This PR aims at fixing a bug that was hidden inside the `AbstractRawDataset` as a result of a transition from the old data management and the new data management methodologies of HydraGNN. 
`update_predicted_values` and `update_atom_features` removed from `AbstractRawDataset` class because these functions now are already included in the `SimplePickleDataset` class.

pzhanggit (commented):
Looks good to me. One comment about the feature update lines is that if we should keep them here and remove the ones in `SimplePickleDataset` since the data is loaded here?

allaffa:
> Looks good to me. One comment about the feature update lines is that if we should keep them here and remove the ones in `SimplePickleDataset` since the data is loaded here?

@pzhanggit that's a good point. AbstractRawDataset is meant to be an "intermediate" layer between the AbstractBaseDataset class and any fully specified dataset class. 
Therefore, for some specific datasets, as we already do for some examples such as for the **ORNL_AISD-Ex** dataset, we inherit directly from AbstractBaseDataset. 
`SimplePickleDataset `or `AdiosDataset` are always used no matter what, because we will always need to convert the data from raw format into pickle or adios. 

Therefore, to me it seems like keeping the feature update lines into `SimplePickleDataset` and `AdiosDataset` is more robust.

pzhanggit (approved):

allaffa:
@pzhanggit @jychoi-hpc 
I temporarily blocked this PR. 
I looked again at the implementation of `AdiosDataset`. 
I want to talk with Jong. I think that what I did is very practical for the `SimplePickleDataset,` but not so sure for `AdiosDataset`

allaffa:
@jychoi-hpc @pzhanggit 
I gave another thought to this PR. 
Many examples not currently included in the main branch also use `update_predicted_values` and `update_atom_features` .  Because of this, the most practical way to ensure that these two capabilities are always available consists maintaining them inside the `SimplePickleDataset` class.

PR #207, opened 2024-01-16
Log 2023 qm9 example update

frobnitzem:
These changes make the analysis step work with the torch 2.1.1+cu121 version I am using.

A try/except block over training allow loading an existing model rather than starting again from scratch.  It was useful for plotting outputs differently.


pzhanggit (approved):
It reads good to me.

PR #209, opened 2024-02-24
software dependencies updated

allaffa:


jychoi-hpc:
Can we try with the following version set? Those are the currently working set on Frontier:
pytorch_scatter 2.1.2
pytorch_sparse 0.6.18
pytorch_cluster 1.6.3
pytorch_spline_conv 1.2.2
pytorch_geometric 2.5.0

allaffa:
@jychoi-hpc @pzhanggit 

I tried to find new pre-compiled wheels that automatically rely on binary files to combine PyTorch>=2.0 with PyG.
 
All the wheels can be found at this links:
https://data.pyg.org/whl/
 
There are wheels that supposedly make torch-scatter-2.1.2 work with torch 2.1 and torch 2.2.
However, when you try to install them, they all return the error “torch not found”, which seems to be a well-known problem
https://github.com/rusty1s/pytorch_scatter/issues/365
 
The most update wheel that I could make work is the one that  where torch-scatter uses torch<=1.13.
 
Please see PR#209
https://github.com/ORNL/HydraGNN/actions/runs/8031604060/job/21940095437?pr=209
 
I think that we need to maintain a very detail requirement.txt and reuirement-torchdep.txt where we clearly specify all the version of the packages.
 
If we say “package >= X.X”, this may lead to individually install the latest version fo a package, but that version may be incompatible with other packages.

PR #210, opened 2024-02-25
Update software

jychoi-hpc:
WIP: upgrading software


.github/workflows/CI.yml
```diff
@@ -34,12 +34,14 @@ jobs:
...
           path: ${{ env.pythonLocation }}
-          key: ${{ env.pythonLocation }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('requirements-torchdep.txt') }}
+          key: ${{ env.pythonLocation }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-dev.txt') }}-${{ hashFiles('requirements-torch.txt') }}-${{ hashFiles('requirements-torchdep.txt') }}
       - name: Install dependencies
         if: steps.cache.outputs.cache-hit != 'true'
         run: |
           python -m pip install --upgrade pip
+          # python -m pip install torch==2.0.1 torchvision torchaudio --index-url https://download.pytorch.org/whl/cpu
           python -m pip install --upgrade -r requirements.txt -r requirements-dev.txt
+          python -m pip install --upgrade -r requirements-torch.txt
```
allaffa, line 44: Does creating an additional `requirements-torch.txt` actually help enforce the order in which dependencies are installed?

.github/workflows/CI.yml
```diff
@@ -50,5 +52,5 @@ jobs:
           # FIXME: install
           export PYTHONPATH=$PYTHONPATH:~/HydraGNN
           pip list
-          python -m pytest -rfEP -W error -W ignore::DeprecationWarning -W ignore::UserWarning --tb=native
-          mpirun -n 2 --oversubscribe python -m pytest -rfEP -W error -W ignore::DeprecationWarning  -W ignore::UserWarning --tb=native --with-mpi
+          mpirun -n 2 --oversubscribe python -m pytest -rfEP -W ignore::DeprecationWarning  -W ignore::UserWarning --tb=native --with-mpi -x -s
```
allaffa, line 55: Does inverting the order here help avoid the issue with `torch.distributed` calls?

.github/workflows/CI.yml
jychoi-hpc: Try to enforce to install pytorch's official release wheel: `--index-url https://download.pytorch.org/whl/cpu`

.github/workflows/CI.yml
jychoi-hpc: Right. Also, I am trying to debug by limiting cases. It is still work-in-progress.

jychoi-hpc:
I am still puzzeled by the following error. I often get the error, not always but from time to time:
```
AttributeError: module 'torch_geometric.nn.conv.cg_conv_CGConv_propagate' has no attribute 'propagate'
```
@pzhanggit @allaffa have you seen such error before?

jychoi-hpc:
I think I found an issue with pyg 2.5.0: https://github.com/pyg-team/pytorch_geometric/issues/8994 
I am downgrading to pyg 2.3.1.

jychoi-hpc:
I think it is working now. @allaffa and @pzhanggit, can you look?

.github/workflows/CI.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python: [3.7, 3.8]
+        python: ["3.8", "3.9"]
```
allaffa, line 18: Till yesterday, you had python 3.10 as third version of python to try out, and that version of python was making the tests fail. 
Now you remove Python 3.10 from the version of python to test. 
Does this mean that the current combination of installation requirements cannot make HydragNN work on python 3.10?

allaffa (commented):
Clarification needed for Python 3.10

allaffa:
> I am still puzzeled by the following error. I often get this error not always but from time to time:
> 
> ```
>   File "/workspace/hydragnn/run_training.py", line 75, in _
>     model = create_model_config(
>   File "/workspace/hydragnn/models/create.py", line 36, in create_model_config
>     return create_model(
>   File "/workspace/hydragnn/models/create.py", line 189, in create_model
>     model = CGCNNStack(
>   File "/workspace/hydragnn/models/CGCNNStack.py", line 34, in __init__
>     super().__init__(
>   File "/workspace/hydragnn/models/Base.py", line 102, in __init__
>     self._init_conv()
>   File "/workspace/hydragnn/models/Base.py", line 110, in _init_conv
>     self.graph_convs.append(self.get_conv(self.input_dim, self.hidden_dim))
>   File "/workspace/hydragnn/models/CGCNNStack.py", line 43, in get_conv
>     cgcnn = CGConv(
>   File "/usr/local/lib/python3.8/dist-packages/torch_geometric/nn/conv/cg_conv.py", line 57, in __init__
>     super().__init__(aggr=aggr, **kwargs)
>   File "/usr/local/lib/python3.8/dist-packages/torch_geometric/nn/conv/message_passing.py", line 190, in __init__
>     self.__class__._jinja_propagate = module.propagate
> AttributeError: module 'torch_geometric.nn.conv.cg_conv_CGConv_propagate' has no attribute 'propagate'
> ```
> 
> @pzhanggit @allaffa have you seen such error before?

@jychoi-hpc 
Yesterday, inspecting the output of the CI workflow tests, I noticed that you were obtaining this error when you were trying to use `Python 3.10`. However, it seems weird that the version of Python causes this error. This seems more of an error due to the fact that the function method is not found in the implementation of the class, and this should not change with the version of Python used.

jychoi-hpc:
I found the error can happen for any python version. It is from pyg 2.5.0 (https://github.com/pyg-team/pytorch_geometric/issues/8994). After using pyg 2.3.0, it works now. I put back "3.10" python. Let's test with python 3.8 and 3.10. We can add more but it will slow down the ci test workflow.

requirements-torch.txt
```diff
@@ -0,0 +1,5 @@
+--index-url https://download.pytorch.org/whl/cpu
```
pzhanggit, line 1: Would this install cpu version torch only? If so, why do we add this constraint?

hydragnn/utils/distributed.py
```diff
@@ -110,6 +110,10 @@ def get_comm_size_and_rank():
 def setup_ddp():
     """ "Initialize DDP"""
 
+    if dist.is_initialized():
```
pzhanggit, line 113: Oops, did we  init ddp somewhere else in examples?

pzhanggit (commented):
Looks good to me.

requirements-torch.txt
jychoi-hpc: Ah, you are right. I was thinking only CI test.

hydragnn/utils/distributed.py
jychoi-hpc: It is called multiple times, especially, in `tests/test_graphs.py`

jychoi-hpc: This update will ensure that ddp initialized called only once.

requirements-torch.txt
jychoi-hpc: I changed. Can you look at again?

.github/workflows/CI.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python: [3.7, 3.8]
+        python: ["3.8", "3.10"]
```
allaffa, line 18: In my opinion, having also Python 3.9 is important because this is the version that currently runs on OLCF-Frontier. 
I would propose the following:
1. use three versions of python: 3.8, 3.9, 3.X
2. use two versions of python: 3.9, 3.X
Moreover, what should X be? 10, 11, or the latest version, which currently is 12?

allaffa (commented):
@pzhanggit @jychoi-hpc 
Let's discuss about the versions of python to use.
Please see comments that I added with my last review.

.github/workflows/CI.yml
jychoi-hpc: How about using 3.9 and 3.11? Those are the python versions on Frontier. 3.12 looks too new.

allaffa: @jychoi-hpc 

Yes. Let’s use 3.9 and 3.11

allaffa (commented):
Yes. Let’s use Python 3.9 and 3.11

requirements.txt
```diff
@@ -1,7 +1,6 @@
 click==8.0.0
 pickle5
```
allaffa, line 2: @jychoi-hpc 

With Python versons >= 3.9, pickle is not needed anymore.

allaffa (commented):
Pickle is not needed anymore for Python versions >= 3.9

allaffa (approved):
Looks fine to me.

PR #214, opened 2024-03-25
Open catalyst 2022

allaffa:
This PR include the following:

1. removes unused import in Open Catalyst 2020 example
2. includes new example for Open Catalyst 2022. I tested the code on Frontier to create AIOS binary files and kick off the training, and it works.

### DIFF
--- a/examples/open_catalyst_2020/open_catalyst_energy.json
+++ b/examples/open_catalyst_2020/open_catalyst_energy.json
@@ -18,7 +18,7 @@
             "num_radial": 6,
             "num_spherical": 7,
             "num_filters": 126,
-            "edge_features": ["coord_x", "coord_y", "coord_z"],
+            "edge_features": ["length"],
             "hidden_dim": 50,
             "num_conv_layers": 3,
             "output_heads": {
--- a/examples/open_catalyst_2020/open_catalyst_forces.json
+++ b/examples/open_catalyst_2020/open_catalyst_forces.json
@@ -18,7 +18,7 @@
             "num_radial": 6,
             "num_spherical": 7,
             "num_filters": 126,
-            "edge_features": ["coord_x", "coord_y", "coord_z"],
+            "edge_features": ["length"],
             "hidden_dim": 50,
             "num_conv_layers": 3,
             "output_heads": {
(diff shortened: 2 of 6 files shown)


jychoi-hpc (approved):
It looks great. Thanks.

pzhanggit (commented):
Looks good to me overall.
One comment is that in the current implementation, using the relative node position as an edge feature does not seem to achieve rotation equivariance.

allaffa:
> Looks good to me overall. One comment is that in the current implementation, using the relative node position as an edge feature does not seem to achieve rotation equivariance.

@pzhanggit 
Goo point. In preparation for the GB work, I do not know whether it is best to just use bond length and use equivariance, or renounce to equivariance and add more informative edge features. 
Honestly, I do not even think we will have time to test all the possible combinations of input parameters to see what works best. 
What do you suggest?

pzhanggit:
> > Looks good to me overall. One comment is that in the current implementation, using the relative node position as an edge feature does not seem to achieve rotation equivariance.
> 
> @pzhanggit Goo point. In preparation for the GB work, I do not know whether it is best to just use bond length and use equivariance, or renounce to equivariance and add more informative edge features. Honestly, I do not even think we will have time to test all the possible combinations of input parameters to see what works best. What do you suggest?

Given the short timeline, I suggest we remove the relative position from edge feature to keep the equivariance.

allaffa:
> > > Looks good to me overall. One comment is that in the current implementation, using the relative node position as an edge feature does not seem to achieve rotation equivariance.
> > 
> > 
> > @pzhanggit Goo point. In preparation for the GB work, I do not know whether it is best to just use bond length and use equivariance, or renounce to equivariance and add more informative edge features. Honestly, I do not even think we will have time to test all the possible combinations of input parameters to see what works best. What do you suggest?
> 
> Given the short timeline, I suggest we remove the relative position from edge feature to keep the equivariance.

@pzhanggit 
I just updated both Open Catalyst 2020 and Open Catalyst 2022 examples so that only the distance between nodes is used as edge feature.

pzhanggit (approved):
Looks good to me.

PR #216, opened 2024-03-26
Qm7x dataset

allaffa:
@frobnitzem 
among all the energies that can be retrieved from the QM7x dataset, may you help me identify the one that corresponds to the energy returned by OC2020 and OC2022?

### DIFF
--- a/examples/qm7x/inference.py
+++ b/examples/qm7x/inference.py
@@ -39,6 +39,7 @@ except ImportError:
     pass
 
 import matplotlib.pyplot as plt
+
 plt.rcParams.update({"font.size": 16})
 
 
@@ -71,12 +72,11 @@ def get_log_name_config(config):
         )
     )
 
+
 def getcolordensity(xdata, ydata):
     ###############################
     nbin = 20
-    hist2d, xbins_edge, ybins_edge = np.histogram2d(
-        x=xdata, y=ydata, bins=[nbin, nbin]
-    )
+    hist2d, xbins_edge, ybins_edge = np.histogram2d(x=xdata, y=ydata, bins=[nbin, nbin])
     xbin_cen = 0.5 * (xbins_edge[0:-1] + xbins_edge[1:])
     ybin_cen = 0.5 * (ybins_edge[0:-1] + ybins_edge[1:])
     BCTY, BCTX = np.meshgrid(ybin_cen, xbin_cen)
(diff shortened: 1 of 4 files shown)


frobnitzem:
@allaffa Comparing energies is complicated because QM methods have many settings which modify their energy function:
- density functional (e.g. LDA, PBE or PBE0 or HSE06 or B3LYP)
- pseudopotential (e.g Ultrasoft, Norm-Conserving, or PAW) 
- "U" parameter applied to metal d- or f- electrons
- basis set / planewave cutoff

And there are 2 different ways to reference energies between different QM methods (to make comparison of the "energy" across methods valid).  Both subtract off the energy of all isolated atoms computed with the same QM method.  One (simple) way treats isolated atoms as single, uncharged atoms in gas phase.  The other (physical) way is to treat isolated atoms as part of their equilibrium material (e.g. N2 gas for nitrogen, solid graphite for carbon, liquid for bromine, etc).

Among the [QM7-x properties](https://zenodo.org/records/4288677), the most useful energies for comparison are:
- 'ePBE0': PBE0 energy [eV] (1)
- 'eAT': PBE0 atomization energy [eV] (1)

The first one is the total energy output by DFT with the PBE0 functional.  The second one is the PBE0 energy after subtracting the energies of the isolated atoms using the first (simple) method.  Ideally we should fit that second one.
see [atomic energies in Table S3](https://www.nature.com/articles/s41597-021-00812-2#MOESM1)

OpenCatalyst provides "raw DFT" energies output from VASP with DFT+U as well as (since OC22) reference energies using the second (physical) method:  https://discuss.opencatalystproject.org/t/about-the-reference-energy/139/7

So we need to adjust the referenced energies of OC or QM7x to compare.  Despite being non-physical, I favor the atomization energies because the QM calculations are more exact (single atom vs. liquid, for example).

allaffa:
@frobnitzem 
Thank you for the explanation. 
For the moment, I would prefer to choose the value of the energy that allows consistency across datasets, and that allows us to ensure that this consistency check is completed before the submission of the GB paper to SC24. 
Given the time constraint, do you think that computing the simple atomization energies for every atomistic structure is something that can be done before April 15?

allaffa:
@pzhanggit @jychoi-hpc 
This is ready for review.

jychoi-hpc (approved):
It looks good to me.

PR #218, opened 2024-03-29
Testing hpo packages

allaffa:
This PR introduces three main examples for hyperparameter optimization (HPO) using the QM9 dataset:

1. `qm9_optuna.py`: HPO example using `optuna`
2. `qm9_deephyper.py`: HPO example running in sequential using `deephyper`
3. `qm9_deephyper_multi.py`: HPO example running in parallel using `deephyper`

### DIFF
--- a/examples/qm9_hpo/qm9.py
+++ b/examples/qm9_hpo/qm9.py
@@ -112,3 +112,6 @@ hydragnn.train.train_validate_test(
     log_name,
     verbosity,
 )
+
+hydragnn.utils.save_model(model, optimizer, log_name)
+hydragnn.utils.print_timers(verbosity)
--- a/examples/qm9_hpo/qm9.py
+++ b/examples/qm9_hpo/qm9.py
@@ -1,0 +1,114 @@
+import os, json
+
+import torch
+import torch_geometric
+
+# deprecated in torch_geometric 2.0
+try:
+    from torch_geometric.loader import DataLoader
+except:
+    from torch_geometric.data import DataLoader
+
+import hydragnn
+import argparse
+
+# Update each sample prior to loading.
+def qm9_pre_transform(data):
+    # Set descriptor as element type.
+    data.x = data.z.float().view(-1, 1)
+    # Only predict free energy (index 10 of 19 properties) for this run.
+    data.y = data.y[:, 10] / len(data.x)
+    graph_features_dim = [1]
+    node_feature_dim = [1]
+    return data
+
+
+def qm9_pre_filter(data):
+    return data.idx < num_samples
+
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--model_type", help="model_type", default="GIN")
+parser.add_argument("--hidden_dim", type=int, help="hidden_dim", default=5)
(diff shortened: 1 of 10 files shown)


jychoi-hpc:
Can we save this qm9 as a separate directory (say, qm9_hpo)? We  often use qm9 as a minimum case. It would be great if we make "qm9" as simple as possible. How do you think?

allaffa:
> Can we save this qm9 as a separate directory (say, qm9_hpo)? We often use qm9 as a minimum case. It would be great if we make "qm9" as simple as possible. How do you think?

@jychoi-hpc agree.

jychoi-hpc (approved):
Thank you for the update. It looks good to me.

jychoi-hpc (approved):
It looks good to me. Thanks.

examples/qm9_hpo/qm9_deephyper_multi.py
```diff
@@ -0,0 +1,189 @@
...
+    # NOTE: data is moved to the device in the pre-transform.
+    # NOTE: transforms/filters will NOT be re-run unless the qm9/processed/ directory is removed.
+    dataset = torch_geometric.datasets.QM9(
+        root="dataset/qm9", pre_transform=qm9_pre_transform
+    )
+
+    trainset, valset, testset = hydragnn.preprocess.split_dataset(dataset, 0.8, False)
+    (train_loader, val_loader, test_loader) = hydragnn.preprocess.create_dataloaders(
+        trainset, valset, testset, 64
+    )
```
pzhanggit, line 134: Could we pass the data loaders defined here as command line arguments in `qm9.py` if we want each trail to reuse the same loaders?

examples/qm9_hpo/qm9.py
```diff
@@ -0,0 +1,114 @@
...
+    optimizer,
+    train_loader,
+    val_loader,
+    test_loader,
+    writer,
+    scheduler,
+    config["NeuralNetwork"],
+    log_name,
+    verbosity,
+)
```
pzhanggit, line 114: `save_model` after training

pzhanggit (commented):
The PR looks good to me overall. I'll need to learn more about these hpo tools. Three general comments: 1) remove or clean up unused functions in `dh_utils.py`; 2) save trained models in all trials in `qm9.py`; 3) maybe add data loaders as command line arguments to avoid repeated data preprocessing in trails.

examples/qm9_hpo/qm9_deephyper_multi.py
allaffa: For large scale runs, where we data loaders are split across processes, I do not see a practical way to do this.

examples/qm9_hpo/qm9.py
allaffa: Thanks. I added these lines.

allaffa:
@pzhanggit 
I also added num_headlayers and dim_headlayers as hyper-parameters

jychoi-hpc (approved):
Great. It looks good to me.

PR #219, opened 2024-03-30
Ani1x

allaffa:
In this PR, I did the following:

- added option to normalize the potential energy by the number of atoms in OC2020, OC2022, MPTrj examples
- removed useless arguments from MPTrj example
- added ANI1x as additional example
- added option to normalize the potential energy for ANI1x example


jychoi-hpc (approved):
Thank you for the PR. It looks good to me.

PR #221, opened 2024-04-01
Add multi dataset option for GB24

jychoi-hpc:
Two update:
* using multiple datasets for GB24
* addiiotn HYDRAGNN_USE_VARIABLE_GRAPH_SIZE env to skip the check if graph size is variable or not at starting.

### DIFF
--- a/examples/multidataset/train.py
+++ b/examples/multidataset/train.py
@@ -112,13 +112,13 @@ if __name__ == "__main__":
         datefmt="%H:%M:%S",
     )
 
-    log_name = "OC2020" if args.log is None else args.log
+    log_name = "GFM" if args.log is None else args.log
     hydragnn.utils.setup_log(log_name)
     writer = hydragnn.utils.get_summary_writer(log_name)
 
     log("Command: {0}\n".format(" ".join([x for x in sys.argv])), rank=0)
 
-    modelname = "OC2020" if args.modelname is None else args.modelname
+    modelname = "GFM" if args.modelname is None else args.modelname
 
     tr.initialize()
     tr.disable()
--- a/examples/multidataset/train.py
+++ b/examples/multidataset/train.py
@@ -1,41 +1,26 @@
-import os, re, json
+import os, json
 import logging
 import sys
 from mpi4py import MPI
 import argparse
 
-import glob
-
-import random
-import numpy as np
-
 import torch
-from torch import tensor
-from torch_geometric.data import Data
 
 import hydragnn
 from hydragnn.utils.time_utils import Timer
 from hydragnn.utils.model import print_model
-from hydragnn.utils.abstractbasedataset import AbstractBaseDataset
 from hydragnn.utils.distdataset import DistDataset
-from hydragnn.utils.pickledataset import SimplePickleWriter, SimplePickleDataset
-from hydragnn.preprocess.utils import gather_deg
-from hydragnn.preprocess.load_data import split_dataset
+from hydragnn.utils.pickledataset import SimplePickleDataset
 
 import hydragnn.utils.tracer as tr
 
-from hydragnn.utils.print_utils import iterate_tqdm, log
-
-from utils.atoms_to_graphs import AtomsToGraphs
-from utils.preprocess import write_images_to_adios
+from hydragnn.utils.print_utils import log
+from hydragnn.utils import nsplit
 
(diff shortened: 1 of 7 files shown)


allaffa (changes requested):
I think it would be better if we create a standalone example to pull multiple datasets together.

examples/multidataset/train.py
```diff
@@ -0,0 +1,282 @@
...
+    comm = MPI.COMM_WORLD
+
+    ## Set up logging
+    logging.basicConfig(
+        level=logging.INFO,
+        format="%%(levelname)s (rank %d): %%(message)s" % (rank),
+        datefmt="%H:%M:%S",
+    )
+
+    log_name = "OC2020" if args.log is None else args.log
```
allaffa, line 115: Let's change the name of the logdir to be somethign more indicative of the examble. 
We can say GFM for (graph foundational model), or GB24, or SC24... your pick :)

examples/multidataset/train.py
```diff
@@ -0,0 +1,282 @@
...
+    hydragnn.utils.setup_log(log_name)
+    writer = hydragnn.utils.get_summary_writer(log_name)
+
+    log("Command: {0}\n".format(" ".join([x for x in sys.argv])), rank=0)
+
+    modelname = "OC2020" if args.modelname is None else args.modelname
```
allaffa, line 121: Let's change the name of the modelname to be something more indicative of the example.
We can say GFM for (graph foundational model), or GB24, or SC24... your pick :)

examples/multidataset/train.py
```diff
@@ -0,0 +1,282 @@
...
+        modellist = args.multi_model_list.split(",")
+        processlist = list(map(lambda x: int(x), args.multi_process_list.split(",")))
+        assert comm_size == sum(processlist)
+        colorlist = list()
+        color = 0
+        for n in processlist:
+            for _ in range(n):
+                colorlist.append(color)
+            color += 1
+        mycolor = colorlist[rank]
```
allaffa, line 180: To my understanding, the coloring assigns different colors to graph samples of different datasets. Is that correct?

hydragnn/utils/config_utils.py
```diff
@@ -23,9 +23,13 @@
 def update_config(config, train_loader, val_loader, test_loader):
     """check if config input consistent and update config with model and datasets"""
 
-    graph_size_variable = check_if_graph_size_variable(
-        train_loader, val_loader, test_loader
-    )
+    graph_size_variable = os.getenv("HYDRAGNN_USE_VARIABLE_GRAPH_SIZE")
```
allaffa, line 26: Is this to avoid useless iterations over the dataset in case you know already that the graphs are of variable size?

examples/multidataset/train.py
```diff
@@ -0,0 +1,282 @@
...
+            mycolor,
+            mymodel,
+            len(trainset),
+            len(valset),
+            len(testset),
+        )
+
+        assert not (args.shmem and args.ddstore), "Cannot use both ddstore and shmem"
+        if args.ddstore:
+            opt = {"ddstore_width": args.ddstore_width, "local": True}
```
allaffa, line 208: Why do we use `local=True? `
Is it because we want different ranks to import different datasets?

examples/multidataset/train.py
jychoi-hpc: GFM sounds good.

examples/multidataset/train.py
jychoi-hpc: Assign different color for different datasets. If there are 3 multi datasets, "color" will go from 0, 1, 2. Each process will belong to one of them.

hydragnn/utils/config_utils.py
jychoi-hpc: You are right.

allaffa (commented):
Some clarifications needed :)

examples/multidataset/train.py
jychoi-hpc: It is intended to indicate that the dataset will setup for DDStore is a set of "locally" owned and thus no need to split internally. Previously, we provide a global view to DDStore and DDstore split them.

allaffa (commented):
One more question.

allaffa (approved):

PR #223, opened 2024-04-03
train and HPO python scripts added for GFM testcase

allaffa:
- Removed unused imports from QM9 example
- Introduced iterative objects that spans all output heads in `config` dictionary to change `num_headlayers` and `dim_headlayers` consistently across all heads (both for graph and node types)
- Introduced example that combines multidatasets for GFM and HPO


examples/multidataset_hpo/job-frontier.sh
```diff
@@ -0,0 +1,73 @@
...
+# Safe sleep to let everything start
+sleep 5
+
+echo "Doing something"
+
+# Launch DeepHyper (1 rank per node, NTOT_DEEPHYPER_RANKS <= NNODES here)
+# meaning NGPUS_PER_TRAINING >= 8
+#$NTOT_DEEPHYPER_RANKS 
+#srun -n1 python qm9_deephyper_multi.py
+python qm9_deephyper_multi.py
```
pzhanggit, line 73: Update the run command for `multidataset_hpo` example with `train_hpo.py`

examples/multidataset_hpo/train_hpo.py
```diff
@@ -0,0 +1,157 @@
...
+    # Choose the sampler (e.g., TPESampler or RandomSampler)
+    from deephyper.evaluator import Evaluator, ProcessPoolEvaluator, queued
+    from deephyper.problem import HpProblem
+    from deephyper.search.hps import CBO
+    from hydragnn.utils.deephyper import read_node_list
+
+    # define the variable you want to optimize
+    problem = HpProblem()
+
+    # Define the search space for hyperparameters
```
pzhanggit, line 112: Want to confirm that we are doing HPO for model architectures only for now?

allaffa: Yes. Since we will very likely be able to kick off only 36 concurrent HPO trials (assuming the entire Frontier is available), I think we should prioritize the optimization of architectural parameters. 
What do you think?

examples/multidataset_hpo/job-frontier.sh
allaffa: Thanks.

examples/multidataset_hpo/gfm.py
```diff
@@ -0,0 +1,241 @@
...
+info(
+    "trainset,valset,testset size: %d %d %d"
+    % (len(trainset), len(valset), len(testset))
+)
+
+if args.ddstore:
+    os.environ["HYDRAGNN_AGGR_BACKEND"] = "mpi"
+    os.environ["HYDRAGNN_USE_ddstore"] = "1"
+
+train, val, test = hydragnn.preprocess.split_dataset(
```
pzhanggit, line 199: Not needed anymore since trainset, valset, and testset are already there?

examples/qm9_hpo/qm9_deephyper.py
```diff
@@ -26,13 +26,6 @@ def qm9_pre_transform(data):
...
-# Set this path for output.
-try:
-    os.environ["SERIALIZED_DATA_PATH"]
-except:
-    os.environ["SERIALIZED_DATA_PATH"] = os.getcwd()
-
-
 def run(trial):
 
     global config, log_name, train_loader, val_loader, test_loader
```
pzhanggit, line 31: What does `global` mean here? Would the variables be shared across all trials?

allaffa: Yes. Here `global ` ensures that the variables are picked from the external "scope" of the function. This trick applies here because this script is an example where each trial is executed in sequential. Unfortunately, this trick does not apply for our large scale problems, because there each trial is kicked off by an independent `srun`, and `global` variables do not survive across multiple `sruns`.

examples/multidataset_hpo/gfm.py
allaffa: Thanks :) You are right. 

I also noticed that `train`, `val`, and `test` in the following line should be `trainset`, `valet`, and `testset`.

pzhanggit (approved):
Looks good to me, thanks

PR #225, opened 2024-04-04
Multi hpo case using deephyper

jychoi-hpc:


hydragnn/train/train_validate_test.py
```diff
@@ -408,10 +411,16 @@ def train(
         and hasattr(loader.dataset.ddstore, "epoch_begin")
         and bool(int(os.getenv("HYDRAGNN_USE_ddstore", "0")))
     )
+    extra = 0 if loader.drop_last else 1
+    nbatch = len(loader.sampler) // loader.batch_size + extra
+
+    if os.getenv("HYDRAGNN_MAX_NUM_BATCH") is not None:
```
allaffa, line 417: Why do we need to set up an environment veriable for the batches?

allaffa (commented):
Some clarifications needed

hydragnn/train/train_validate_test.py
jychoi-hpc: This is to limit the number of batches. For example, in HPO, we want to run only 100 batches instead of, say, 1000 batches per epoch in the full data. This is a temporary solution for now.

jychoi-hpc: Similarly, I put HYDRAGNN_VALTEST env too, which is to skip validation and test.

examples/multidataset_hpo/gfm_deephyper_multi.py
```diff
@@ -57,7 +59,10 @@ def run(trial, dequed=None):
             f"--ntasks-per-node=8 --gpus-per-node=8",
             f"--cpus-per-task {OMP_NUM_THREADS} --threads-per-core 1 --cpu-bind threads",
             f"--gpus-per-task=1 --gpu-bind=closest",
-            f"--export=ALL,{master_addr}",
+            f"--export=ALL,{master_addr},HYDRAGNN_VALTEST=1,HYDRAGNN_MAX_NUM_BATCH=100,HYDRAGNN_USE_VARIABLE_GRAPH_SIZE=1,HYDRAGNN_AGGR_BACKEND=mpi",
```
allaffa, line 62: should we remove` HYDRAGNN_VALTEST=1` ?

allaffa (commented):
Question on environment variable HYDRAGNN_VALTEST=1

examples/multidataset_hpo/gfm_deephyper_multi.py
jychoi-hpc: yes.

PR #226, opened 2024-04-05
Remove multi_process_list option in multidataset

jychoi-hpc:
Remove multi_process_list option in multidataset. It calculates number of processes per dataset based on the number of molecules in the files.

### DIFF
--- a/examples/multidataset/train.py
+++ b/examples/multidataset/train.py
@@ -5,6 +5,7 @@ from mpi4py import MPI
 import argparse
 
 import torch
+import numpy as np
 
 import hydragnn
 from hydragnn.utils.time_utils import Timer
@@ -47,9 +48,6 @@ if __name__ == "__main__":
     parser.add_argument("--modelname", help="model name")
     parser.add_argument(
         "--multi_model_list", help="multidataset list", default="OC2020"
-    )
-    parser.add_argument(
-        "--multi_process_list", help="multidataset process list", default="1"
     )
 
     group = parser.add_mutually_exclusive_group()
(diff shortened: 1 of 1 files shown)


examples/multidataset/train.py
```diff
@@ -171,13 +169,31 @@ def info(*args, logtype="info", sep=" "):
...
+            ndata_list = np.array(ndata_list, dtype=np.float32)
+            process_list = np.ceil(ndata_list / sum(ndata_list) * comm_size).astype(
+                np.int32
+            )
+            imax = np.argmax(process_list)
+            process_list[imax] = process_list[imax] - (np.sum(process_list) - comm_size)
+            process_list = process_list.tolist()
+            comm.bcast(process_list)
+        else:
+            process_list = comm.bcast(None)
```
allaffa, line 191: why do processes with `rank>0` need to broadcast `None` ?

allaffa (commented):
One clarification needed.

examples/multidataset/train.py
jychoi-hpc: They only need to receive data which is the left side of "=". Rank 0, in the mean time, need to send the original data which is the input in the parenthesis.

allaffa: I would have done the following:
 
```
if rank = 0:
 
      ....
    process_list = process_list.tolist()

#exit the if-statement
process_list = comm.bcast(process_list, root=0)
```

is what you wrote equivalent to what I would have done?

allaffa (commented):
One more clarification needed :)

examples/multidataset/train.py
```diff
@@ -171,13 +169,29 @@ def info(*args, logtype="info", sep=" "):
...
+                ndata_list.append(f.ndata)
+                del f
+            ndata_list = np.array(ndata_list, dtype=np.float32)
+            process_list = np.ceil(ndata_list / sum(ndata_list) * comm_size).astype(
+                np.int32
+            )
+            imax = np.argmax(process_list)
+            process_list[imax] = process_list[imax] - (np.sum(process_list) - comm_size)
+            process_list = process_list.tolist()
+        process_list = comm.bcast(process_list)
```
allaffa, line 189: shouldn't you specify the root? 

`process_list = comm.bcast(process_list, **root=0**)`

allaffa (commented):
Last question. I promise.

allaffa (approved):

examples/multidataset/train.py
```diff
@@ -171,13 +169,29 @@ def info(*args, logtype="info", sep=" "):
...
+                f = AdiosDataset(fname, "trainset", comm)
+                ndata_list.append(f.ndata)
+                del f
+            ndata_list = np.array(ndata_list, dtype=np.float32)
+            process_list = np.ceil(ndata_list / sum(ndata_list) * comm_size).astype(
+                np.int32
+            )
+            imax = np.argmax(process_list)
+            process_list[imax] = process_list[imax] - (np.sum(process_list) - comm_size)
+            process_list = process_list.tolist()
```
allaffa, line 188:
```suggestion
            process_list = process_list.tolist()
       else:
            process_list = None
```

allaffa (changes requested):
One suggestion. Otherwise processes with rank > 0 do not have a memory buffer where to receive `process_list ` coming from rank = 0

examples/multidataset/train.py
```diff
@@ -171,13 +169,29 @@ def info(*args, logtype="info", sep=" "):
...
         modellist = args.multi_model_list.split(",")
-        processlist = list(map(lambda x: int(x), args.multi_process_list.split(",")))
-        assert comm_size == sum(processlist)
+        if rank == 0:
+            ndata_list = list()
+            for model in modellist:
+                fname = os.path.join(
+                    os.path.dirname(__file__), "./dataset/%s.bp" % model
+                )
+                f = AdiosDataset(fname, "trainset", comm)
```
pzhanggit, line 179: Is rank 0 reading the entire Adios data file here? If so, would it be too big and could we just ask it to get the length since all we need is `f.ndata`?

jychoi-hpc: No, rank=0 only opens and gets the length. No data reading.

examples/multidataset/train.py
jychoi-hpc: You are right

allaffa (approved):
Looks fine to me.

PR #229, opened 2024-04-06
pyg_lib corrected into pyg-lib

allaffa:
import of "pyg-lib" corrected in the` requirements.txt`


jychoi-hpc (approved):
Great. Thank you for the fix.

PR #238, opened 2024-04-13
flipped sign in validation loss functions return by HPO trials

allaffa:
flipped sign in validation loss functions because DeepHyper maximize the objective function


jychoi-hpc (approved):
Great. It looks good.

PR #244, opened 2024-05-06
bug fixed

allaffa:


allaffa:
@jychoi-hpc 
Previously, I had only created ADIOS binary files without kicking off the training. 
Now I solved bug fixed to make the training work, and it runs on OLCF-Frontier with ADIOS+DDStore

PR #245, opened 2024-05-07
Minor updates on DDStore and remaining time checking

jychoi-hpc:
This includes a few updates:
* Update on DDStore epoch begin/end routine 
* Adding check_remaining time routine within a job allocation on Frontier and Perlmutter

### DIFF
--- a/hydragnn/utils/distributed.py
+++ b/hydragnn/utils/distributed.py
@@ -292,13 +292,16 @@ def check_remaining(t0):
     device = get_device()
     if jobid is not None:
         if world_rank == 0:
-            cmd = f"squeue -h -j {jobid} -o %L"
-            proc = subprocess.run(cmd.split(), stdout=subprocess.PIPE)
-            timestr = proc.stdout.decode("utf-8").strip()
-            left = timedelta_parse(timestr).total_seconds()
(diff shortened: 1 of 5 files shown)


hydragnn/utils/distributed.py
```diff
@@ -265,3 +267,41 @@ def comm_reduce(x, op):
...
+
+def check_remaining(t0):
+    ## Early stop
+    world_size, world_rank = get_comm_size_and_rank()
+    jobid = os.getenv("SLURM_JOB_ID", None)
+    should_stop = False
+    device = get_device()
+    if jobid is not None:
+        if world_rank == 0:
+            cmd = f"squeue -h -j {jobid} -o %L"
```
kshitij-v-mehta, line 295: Could you add a try-catch block here to catch exceptions that may occur if the slurm command returns an error? This code snippet assumes the system uses Slurm, which may not be true for external users.

allaffa:
@jychoi-hpc 
I think there are black formatting incompatibilities that cause the CI tests to crash

allaffa (approved):

kshitij-v-mehta (approved):

PR #248, opened 2024-05-28
Removed pickle5 requirement from setup.py

erdemcaliskan:
pickle5 is replaced with pickle for python > 3.7, thus pickle5 requirement is not needed.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,6 @@ def read(fname):
 
 
 install_requires = [
-    "pickle5",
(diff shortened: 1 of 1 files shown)


allaffa (approved):

allaffa:
@jychoi-hpc 
Is this also ok to merge?

jychoi-hpc (approved):
Yes, it looks right. Thank you for the fix.

PR #250, opened 2024-06-04
Enhancement: model-agnostic layer-wise checkpointing

LemonAndRabbit:
[Checkpointing](https://pytorch.org/docs/stable/checkpoint.html) is a widely used technique (i.e. in [deepspeed](https://github.com/microsoft/DeepSpeed/blob/f4cb866c2eba60a72d79c3741d12d098fe4e45b8/deepspeed/runtime/activation_checkpointing/checkpointing.py#L725)) to reduce peak GPU memory consumption. It achieves this by recomputing intermediate activations during the backward pass instead of storing them from the forward pass.

This enhancement leverages PyTorch's official checkpoint implementation and applies it at a per convolution layer granuality.

Key features include:

+ Model agnosticism: It can be used with any type of message-passing layers.
+ Significant memory reduction: An 18-layer, 512-dimension PNA network (with batch_size=128) shows a peak memory usage reduction of approximately 58%.

Usage: config["NeuralNetwork"]["Training"]["conv_checkpointing"] = True

Overhead: With an 18-layer, 512-dimension PNA network, recomputing activations introduces an extra training time of ~28.6%, as observed on an 8x A5000 server.

Raw profiling results: Attached are the memory consumption and latency profiling results ([google_drive](https://drive.google.com/drive/folders/1xGRj8AouAFfV_kUnnNDR8AoF-WfaDeFQ?usp=sharing)). For visualization, please refer to [pytorch.org/memory_viz](https://pytorch.org/memory_viz) and [ui.perfetto.dev/](https://ui.perfetto.dev/), respectively.

### DIFF
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -18,6 +18,7 @@ from torch.utils.checkpoint import checkpoint
 from hydragnn.utils.model import activation_function_selection, loss_function_selection
 import sys
 from hydragnn.utils.distributed import get_device
+from hydragnn.utils.print_utils import print_master
 
 import inspect
 
@@ -282,7 +283,7 @@ class Base(Module):
             self.heads_NN.append(head_NN)
 
     def enable_conv_checkpointing(self):
-        print("Enabling checkpointing")
+        print_master("Enabling checkpointing")
         self.conv_checkpointing = True
 
     def forward(self, data):
--- a/hydragnn/models/Base.py
+++ b/hydragnn/models/Base.py
@@ -14,6 +14,7 @@ from torch.nn import ModuleList, Sequential, ReLU, Linear, Module
 import torch.nn.functional as F
 from torch_geometric.nn import global_mean_pool, BatchNorm
 from torch.nn import GaussianNLLLoss
+from torch.utils.checkpoint import checkpoint
 from hydragnn.utils.model import activation_function_selection, loss_function_selection
 import sys
 from hydragnn.utils.distributed import get_device
@@ -107,6 +108,8 @@ class Base(Module):
         self._multihead()
         if self.initial_bias is not None:
             self._set_bias()
+
+        self.conv_checkpointing = False
 
     def _init_conv(self):
         self.graph_convs.append(self.get_conv(self.input_dim, self.hidden_dim))
@@ -278,6 +281,10 @@ class Base(Module):
                 )
             self.heads_NN.append(head_NN)
 
+    def enable_conv_checkpointing(self):
+        print("Enabling checkpointing")
+        self.conv_checkpointing = True
+
     def forward(self, data):
         x = data.x
         pos = data.pos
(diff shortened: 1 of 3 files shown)


hydragnn/models/Base.py
```diff
@@ -278,14 +281,23 @@ def _multihead(self):
...
         pos = data.pos
 
         ### encoder part ####
         conv_args = self._conv_args(data)
         for conv, feat_layer in zip(self.graph_convs, self.feature_layers):
-            c, pos = conv(x=x, pos=pos, **conv_args)
+            if not self.conv_checkpointing:
+                c, pos = conv(x=x, pos=pos, **conv_args)
+            else:
+                c, pos = checkpoint(
```
allaffa, line 298: To me it seems like you are checkpointing the entire message passing layer. Based on our previous discussions, I thought that the checkpointing should involve only the non-linear activation functions. 
Would you mind further explaining why you implemented the checkpointing this way?

allaffa (commented):
Some explanations needed.

hydragnn/models/Base.py
LemonAndRabbit: Of course! This design choice takes both **_performance_** and **_model-agnosticism_** into consideration.

Before delving into the specifics of this design decision, let's clarify the term activation checkpointing used in our discussion. In the machine learning literature, activation refers not only to the output of an activation layer but also to any intermediate tensors created during the forward pass and stored for backward computation. A more detailed explanation can be found [here](https://docs.aws.amazon.com/sagemaker/latest/dg/model-parallel-extended-features-pytorch-activation-checkpointing.html) from AWS.

Regarding my implementation:

+ **Performance (memory reduction)**: Approximately 90% of the per-layer memory usage is attributed to the message passing mechanism rather than the linear layer. This should be included in the checkpointing process, which involves the instantiation of many intermediate tensors (i.e. [1](https://github.com/pyg-team/pytorch_geometric/blob/2.5.2/torch_geometric/nn/conv/pna_conv.py#L185), [2](https://github.com/pyg-team/pytorch_geometric/blob/2.5.2/torch_geometric/nn/conv/pna_conv.py#L169)).
+ **model-agnosticism**: The current implementation, which modifies the forward function of the Base class, makes the checkpointing mechanism agnostic to any message passing layer subclasses. This allows direct scalability to newly added subclasses. Checkpointing only the non-linear functions, however, would require modification on a per-subclass basis, thus not being model-agnostic and necessitating manual effort to scale to future message passing layers.

pzhanggit: Thanks! Did you test the `checkpoint` function? What are the memory reduction and computation overhead like?

hydragnn/models/Base.py
```diff
@@ -278,14 +281,23 @@ def _multihead(self):
                 )
             self.heads_NN.append(head_NN)
 
+    def enable_conv_checkpointing(self):
+        print("Enabling checkpointing")
```
pzhanggit, line 285: Use `print_distributed` from `hydragnn.utils.print_utils`

hydragnn/models/Base.py
LemonAndRabbit: Yes, I have tested the memory reduction and computation overhead, you can refer to the [first comment](https://github.com/ORNL/HydraGNN/pull/250#issue-2334319718)

hydragnn/models/Base.py
LemonAndRabbit: Hi Pei, thank you for pointing out! I have switched to print_master here:
```python
    def enable_conv_checkpointing(self):
        print_master("Enabling checkpointing")
        self.conv_checkpointing = True
```

hydragnn/models/Base.py
pzhanggit: Oops, missed that

hydragnn/models/Base.py
LemonAndRabbit: Thank you Max

pzhanggit (approved):

allaffa (approved):

PR #252, opened 2024-06-08
Qm7x check forces

allaffa:
Check on value of L2-norm of force tensors added to filter out atomistic structures that are not feasible.


frobnitzem (approved):
LGTM.

PR #253, opened 2024-06-08
OC2020 check forces

allaffa:
Functionalities to check on value of atomic forces added to example with Open Catalyst 2020 dataset

### DIFF
--- a/examples/open_catalyst_2020/train.py
+++ b/examples/open_catalyst_2020/train.py
@@ -54,6 +54,9 @@ class OpenCatalystDataset(AbstractBaseDataset):
         self.var_config = var_config
         self.data_path = os.path.join(dirpath, data_type)
         self.energy_per_atom = energy_per_atom
+        
+        # Threshold for atomic forces in eV/angstrom
+        self.forces_norm_threshold = 100.0
 
         self.dist = dist
         if self.dist:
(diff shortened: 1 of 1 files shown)


frobnitzem:
LGTM.

frobnitzem (approved):

allaffa:
@jychoi-hpc 
I am trying to re-generate the ADIOS2 binary file with the changes applied by this PR. However, the code on OLCF-Frontier seems to stall at the saving stage without even completing, even when I reserve 50 nodes for 6 hours. May you please try to run the PR yourself and tell me if you also notice the same behavior?

jychoi-hpc:
I see. I will run from my side and check.

jychoi-hpc:
@allaffa I found a bug in the code. It was introduced after a fix made during the GB run (#241). 

Can you add one line in `hydragnn/utils/adiosdataset.py` as follows and try again?
```
diff --git a/hydragnn/utils/adiosdataset.py b/hydragnn/utils/adiosdataset.py
index 324bb40..79d2214 100644
--- a/hydragnn/utils/adiosdataset.py
+++ b/hydragnn/utils/adiosdataset.py
@@ -107,6 +107,7 @@ class AdiosWriter:
                         break
 
                 for k in keys:
+                    vdim = self.comm.allreduce(0, op=MPI.MAX)
                     shape_list = self.comm.allgather([])
 
                 continue

```

allaffa:
> @allaffa I found a bug in the code. It is an Adios write routine to handle a situation when one process has no data to write but others has. It was introduced after a fix made during the GB run (#241).
> 
> Can you add one line in `hydragnn/utils/adiosdataset.py` as follows and try again?
> 
> ```
> diff --git a/hydragnn/utils/adiosdataset.py b/hydragnn/utils/adiosdataset.py
> index 324bb40..79d2214 100644
> --- a/hydragnn/utils/adiosdataset.py
> +++ b/hydragnn/utils/adiosdataset.py
> @@ -107,6 +107,7 @@ class AdiosWriter:
>                          break
>  
>                  for k in keys:
> +                    vdim = self.comm.allreduce(0, op=MPI.MAX)
>                      shape_list = self.comm.allgather([])
>  
>                  continue
> ```

@jychoi-hpc 
This PR is made with a new branch that stems out of the main. How come that I need to add this changes manually? Shouldn't they already be included in the main if they are bug fixes from the GB?

jychoi-hpc:
This is a new bug caused by #214. The above fix is not added anywhere. You can add the line in your local code. If it works, just commit. It will show up here again.

allaffa:
@jychoi-hpc 
PR#255 (already merged) fixed this!
Thank you :) I think we are good.

PR #259, opened 2024-06-16
energy and force added as attributes to data objects for qm7x dataset

allaffa:


allaffa:
@jychoi-hpc 
Please let me know if the new ADIOS2 binary data contains all the attributes as you need.

jychoi-hpc (approved):

jychoi-hpc (approved):

PR #262, opened 2024-07-02
Bug Fix: Preprocessing and Training with the OGB PCQM4Mv2 Dataset

LemonAndRabbit:
We have identified several issues with the original implementation in [examples/ogb/train_gap.py](https://github.com/ORNL/HydraGNN/blob/main/examples/ogb/train_gap.py) that require modifications:

1. **Incompatibility with the Current OGB PCQM4Mv2 Dataset**: The current version of the [OGB PCQM4Mv2](https://ogb.stanford.edu/docs/lsc/pcqm4mv2/) dataset includes atoms not listed in ogb_node_types and contains entries with empty labels. We skipped these incompatible entries in the preprocessing code.
2. **Failed Instantiation of `AdiosDataset`**: The code currently instantiates `AdiosDataset` with incompatible parameters. The `opt` dictionary should be unpacked before being passed as arguments.
3. **Broadcasting Over 2GB Data with MPI**: The `Adios_writer` class occasionally attempts to broadcast over 2GB of data, exceeding the MPI message count limit. We have implemented a chunk-based broadcasting function to address this issue.

These bug fixes are essential for later integrating our DeepSpeed and pipeline-parallelism implementations, which use the OGB PCQM4Mv2 dataset as an example.

### DIFF
--- a/examples/ogb/train_gap.py
+++ b/examples/ogb/train_gap.py
@@ -11,6 +11,7 @@ from mpi4py import MPI
 from itertools import chain
 import argparse
 import time
+import math
 
 import hydragnn
 from hydragnn.preprocess.load_data import split_dataset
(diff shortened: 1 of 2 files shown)


allaffa (approved):

jychoi-hpc (commented):
Thank you for the fix and improvements. It looks great.

jychoi-hpc (approved):

Issue #205, opened 2023-11-29
ADIOS ImportError is being ignored


kshitij-v-mehta:
[https://github.com/ORNL/HydraGNN/blob/main/hydragnn/utils/adiosdataset.py#L10C1-L13C9](https://github.com/ORNL/HydraGNN/blob/main/hydragnn/utils/adiosdataset.py#L10C1-L13C9)

An ImportError caused by a missing ADIOS installation is not handled. This leads to errors with the ADIOS datasets further in the code.
[https://github.com/ORNL/HydraGNN/blob/main/hydragnn/utils/adiosdataset.py#L300](https://github.com/ORNL/HydraGNN/blob/main/hydragnn/utils/adiosdataset.py#L300)

`NameError: name 'ad2' is not defined`

allaffa:
@kshitij-v-mehta @jychoi-hpc 
Is this issue still relevant or can we close it?

kshitij-v-mehta:
It is still open. We need to add a clear error message when someone tries to use ADIOS when it is not installed.

PR #266, opened 2024-07-18
Cleaner way to fail if ADIOS is not installed. #205

kshitij-v-mehta:
This adds a check to see if ADIOS is installed. It will fail with an ImportError if ADIOS is not found. We don't need to add any more error handling here.

Without this line, we get 
```
NameError: name 'ad2' is not defined
```
which is not intuitive.


jychoi-hpc (approved):
Thank you for the fix. It looks good to me.

allaffa (approved):

PR #267, opened 2024-07-22
Adding PNAPlus Stack

RylieWeaver:
This PR adds the PNAPlus stack I've recently created and experimented with. It replicates DimeNet's use of Bessel Functions and Element-Wise product of "message_ij" in its message propagation. In my intuition, and according to some small experiments, this is more effective than raw concatenation as can be done with edge features. Otherwise, this operates the same as regular PNA.

This is an **Invariant Stack**.


jychoi-hpc:
It looks good to me. @allaffa and @pzhanggit how do you think?

allaffa (approved):
Looks good to me. @pzhanggit what do you think?

RylieWeaver:
@pzhanggit Is this good to merge?

hydragnn/models/PNAPlusStack.py
```diff
@@ -0,0 +1,283 @@
...
+
+from typing import Any, Callable, Dict, List, Optional, Union, Tuple
+
+# Torch
+import torch
+from torch import Tensor
+from torch.utils.data import DataLoader
+import torch.nn.functional as F
+from torch.nn import ModuleList, Sequential, Identity
+from torch_geometric.nn import PNAConv
```
pzhanggit, line 20: Remove this line since using the customized `PNAConv` class in the file? Or if it makes sense to build the customized class on top of PyG version?

RylieWeaver: @pzhanggit You're right that I don't need PNAConv and Identity, so I've taken those out. Does that cover the concern? Let me know if there's anything else.

PR #268, opened 2024-07-22
Enhancement: Deepspeed Launch Template & Now Compatible with GPU Binding

LemonAndRabbit:
As noted by @jychoi-hpc in the [previous pull request](https://github.com/ORNL/HydraGNN/pull/264#issuecomment-2237505680), the DeepSpeed backend fails to initialize when a rank has only one visible GPU (e.g., when launched with --`gpus-per-task=1`).

The root cause is that DeepSpeed defaults to determining GPU ordinals based on the `LOCAL_RANK` environment variable if no `device_rank` is passed, but with one visible GPU per rank, only device ordinal 0 is valid, which makes all ranks except local_rank 0 fails. 

Although this default mapping works when granting visibility of all GPUs to all ranks, it may not be optimal considering the [heterogeneous connectivity](https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#frontier-compute-nodes) between CPU caches and GPUs on Frontier. For optimal performance, it is recommended to use `--gpus-per-task=1 --gpu-bind=closest`, as outlined [here](https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#gpu-mapping).

In this pull request, I have addressed the issue by explicitly specifying which GPU to use (i.e. passing a `device_rank=0` to deepspeed if only having one visible GPU). This ensures that the DeepSpeed backend functions correctly with both one visible GPU per rank (e.g., launched with `--gpus-per-task=1`) and all GPUs visible to all ranks (e.g., launched with `--gres=gpu:8`).

Additionally, I have included a bash template to facilitate running the DeepSpeed backend on Frontier.

### DIFF
--- a/hydragnn/utils/distributed.py
+++ b/hydragnn/utils/distributed.py
@@ -233,13 +233,15 @@ def get_device_name(use_gpu=True, rank_per_model=1, verbosity_level=0, no_prefix
 
     return device_name
 
+
 def get_deepspeed_init_args():
     class Obj(object):
         pass
-    
+
     obj = Obj()
     obj.device_rank = int(get_device_name(no_prefix=True))
     return obj
+
 
 def get_local_rank():
     localrank = 0
--- a/hydragnn/utils/distributed.py
+++ b/hydragnn/utils/distributed.py
@@ -198,7 +198,7 @@ def get_device_list():
     return available_gpus
 
 
-def get_device_name(use_gpu=True, rank_per_model=1, verbosity_level=0):
+def get_device_name(use_gpu=True, rank_per_model=1, verbosity_level=0, no_prefix=False):
 
     available_gpus = get_device_list()
     if not use_gpu or not available_gpus:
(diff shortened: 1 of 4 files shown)


allaffa (commented):
I need some clarifications.

hydragnn/utils/distributed.py
```diff
@@ -226,11 +226,23 @@ def get_device_name(use_gpu=True, rank_per_model=1, verbosity_level=0):
...
 
     return device_name
 
 
+def get_deepspeed_init_args():
+    class Obj(object):
+        pass
+
+    obj = Obj()
+    obj.device_rank = int(get_device_name(no_prefix=True))
```
allaffa, line 242: @LemonAndRabbit 
If I understand correctly, when `gpu-per-rank=1` each rank generates an object with `obj.device_rank=0`. Is this the case?

LemonAndRabbit: Yes, your understanding is correct.

allaffa: > Yes, your understanding is correct.

Why does this work whereas passing "coda:0" for each rank does not?

LemonAndRabbit: Because deepspeed expects an integer device ordinal instead of a string for this field.

allaffa: > Because deepspeed expects an integer device ordinal instead of a string for this field.

In the description of the PR, you said that the previous version of the code was making DeepSpeed work if `--gres=gpu:8`. 
However, `get_device_name` was expected to return a string also in that case. So why returning a string with `--gres=gpu:8` was ok, but returning a string with `--gpus-per-task=1 --gpu-bind=closest` was causing issues?

LemonAndRabbit: In the previous version, we did not pass any device_rank into deepspeed, but relied on its [internal code](https://github.com/microsoft/DeepSpeed/blob/v0.14.1/deepspeed/runtime/engine.py#L977) to decide which device to use, which defaults to `LOCAL_RANK` environmental variable.

allaffa (approved):

pzhanggit (approved):
Looks good to me

PR #270, opened 2024-07-26
add GaussianNLLLoss for UQ

pzhanggit:


hydragnn/models/Base.py
```diff
@@ -308,12 +321,15 @@ def forward(self, data):
...
+        outputs_var = []
         for head_dim, headloc, type_head in zip(
             self.head_dims, self.heads_NN, self.head_type
         ):
             if type_head == "graph":
                 x_graph_head = self.graph_shared(x_graph)
-                outputs.append(headloc(x_graph_head))
+                output_head = headloc(x_graph_head)
+                outputs.append(output_head[:, :head_dim])
+                outputs_var.append(output_head[:, head_dim:] ** 2)
```
allaffa, line 332: Here `output_var` is an NxN matrix, where N is the number of entries in the graph vector output. 
This is because you have to return the entire covariance matrix for graph vector outputs, correct?

hydragnn/models/Base.py
```diff
@@ -253,7 +266,7 @@ def _multihead(self):
                     # """if different graphs in the dataset have different size, one MLP is shared across all nodes """
                     head_NN = MLPNode(
                         self.hidden_dim,
-                        self.head_dims[ihead],
+                        self.head_dims[ihead] * (1 + self.var_output),
```
allaffa, line 269: If the loss function used is NLL, then the output must be two-dimensional in order to return both the mean and the standard deviation, correct?

hydragnn/models/Base.py
```diff
@@ -193,14 +199,21 @@ def _init_node_conv(self):
...
                 self.convs_node_output.append(
-                    self.get_conv(self.hidden_dim_node[-1], self.head_dims[ihead])
+                    self.get_conv(
+                        self.hidden_dim_node[-1],
+                        self.head_dims[ihead] * (1 + self.var_output),
+                    )
                 )
-            self.batch_norms_node_output.append(BatchNorm(self.head_dims[ihead]))
+            self.batch_norms_node_output.append(
+                BatchNorm(self.head_dims[ihead] * (1 + self.var_output))
```
allaffa, line 215: If the loss function used is NLL, then the output must be two-dimensional in order to return both the mean and the standard deviation, correct?

allaffa (commented):
Just a few question to confirm that I understand correctly the proposed changes.

hydragnn/models/Base.py
pzhanggit: `output_var` is the variance on the diagonal and has the same size with the prediction `outputs`.

hydragnn/models/Base.py
pzhanggit: Yes, and we have it done by setting `self.var_output=1` https://github.com/ORNL/HydraGNN/blob/04c167991f36363f93a495241d823c525b10822e/hydragnn/models/Base.py#L74

hydragnn/models/Base.py
pzhanggit: Same comment above

jychoi-hpc (approved):
Great. Thank you for adding this. It looks good to me.

PR #271, opened 2024-07-26
Clintox dataset example

kshitij-v-mehta:


examples/clintox/clintox.json
```diff
@@ -0,0 +1,59 @@
...
+            200
+          ]
+        }
+      },
+      "task_weights": [
+        1
+      ]
+    },
+    "Variables_of_interest": {
+      "input_node_features": [
```
allaffa, line 27: Running the code in debug mode, may you check what is the shape of the tensor `data.x`? For this problem, `data.x` should have only one column. 
If this is the case, then we should use
```
"input_node_features": [
  0
  ],
```

examples/clintox/clintox.json
```diff
@@ -0,0 +1,59 @@
...
+        8,
+        9,
+        10,
+        11
+      ],
+      "output_index": [
+        0
+      ],
+      "type": [
+        "graph"
```
allaffa, line 45: I think it is correct to use only one head to make graph-level predictions.
However, it seems like the predictions require performing a binary classification problem. 
Currently, HydraGNN is set up to perform only regression model. 
To adapt the code for classification, changes are minor, but need to be made. Namely:

- use sigmoid as activation function at the last layer
- use binary cross entropy as loss function for training, validation, and testing

allaffa (commented):
A few comments. Let's discuss it.

kshitij-v-mehta:
Closing this PR, as #272 replaces it

PR #278, opened 2024-09-06
Energy forces

RylieWeaver:
@allaffa Let me know what you think of this PR for energy and force computation. It should follow exactly what we discussed before. Some notable things:

(1) It requires data.energy and data.forces to exist and be named this way
(2) It balances force and energy prediction weighting implicitly, since we aren't able to pass multiple weights in single-task learning
(3) Compute forces is an argument to be specified in the json within "Training".
      - If specified, it throws errors for anything other than nodal single task learning, as well as any model that doesn't
         use positional information for prediction


allaffa (commented):
@RylieWeaver 
This is exactly what we discussed and looks awesome. 
Have you tried it with Lenndard Jones to make sure that it is compatible?

RylieWeaver:
@allaffa 

Awesome, thanks.

I was able to get it to work with Lennard Jones after some small changes to LJ's dataset creation and inference. For example, we now have `data.energy` and`'data.forces` instead of `data.y` and `data.forces_pre_scaled`. A next step could be updating the LennardJones example to that.

Do not merge it yet, because I am resolving any issues with the GitHub tests, but will let you know when they are passed.

allaffa:
> @allaffa 
> 
> 
> 
> Awesome, thanks.
> 
> 
> 
> I was able to get it to work with Lennard Jones after some small changes to LJ's dataset creation and inference. For example, we now have `data.energy` and`'data.forces` instead of `data.y` and `data.forces_pre_scaled`. A next step could be updating the LennardJones example to that.
> 
> 
> 
> Do not merge it yet, because I am resolving any issues with the GitHub tests, but will let you know when they are passed.
> 
> 
> 
> 

@RylieWeaver 
As long as the CI tests do not pass, I would not be allowed to merge this PR anyway. We have explicitly forbidden this option in the settings of the GitHub repository.

allaffa:
@RylieWeaver 

I think it would be good to also include a mew test that the CI workflow must execute where we check this capability to make sure that future changes in the code do not break it

RylieWeaver:
> @RylieWeaver As long as the CI tests do not pass, I would not be allowed to merge this PR anyway. We have explicitly forbidden this option in the settings of the GitHub repository.

Ok, that's good.

RylieWeaver:
> @RylieWeaver
> 
> I think it would be good to also include a mew test that the CI workflow must execute where we check this capability to make sure that future changes in the code do not break it

Ok, I can include this. Any model that includes positional information should be able to predict forces in this method. Should we use that list for the testing?

RylieWeaver:
@allaffa I believe I'll need to create a new dataset inside `hydragnn/serialized_dataset` for this testing, so that data.forces and data.energy are available. It seems like most of the other data is generated randomly from [0.0, 0.5, 1.0].

Is this correct and is random generation is fine for the forces/energy as well?

allaffa:
> @allaffa I believe I'll need to create a new dataset inside `hydragnn/serialized_dataset` for this testing, so that data.forces and data.energy are available. It seems like most of the other data is generated randomly from [0.0, 0.5, 1.0].
> 
> 
> 
> Is this correct and is random generation is fine for the forces/energy as well?

@RylieWeaver 
I have an idea that I would like to run by you. 
What about adding a frozen version of the LJ code as an additional example, where the example first generates the data, saves it into files, and then uses it for training? We don't need to generate many data samples and traing it fully. We just need to make sure that the training kicks off correctly without errors.

RylieWeaver:
> > @allaffa I believe I'll need to create a new dataset inside `hydragnn/serialized_dataset` for this testing, so that data.forces and data.energy are available. It seems like most of the other data is generated randomly from [0.0, 0.5, 1.0].
> > Is this correct and is random generation is fine for the forces/energy as well?
> 
> @RylieWeaver I have an idea that I would like to run by you. What about adding a frozen version of the LJ code as an additional example, where the example first generates the data, saves it into files, and then uses it for training? We don't need to generate many data samples and traing it fully. We just need to make sure that the training kicks off correctly without errors.

@allaffa Yeah, I think this will be good to add LJ to the main branch. Would you be free to hammer out the specifics? Your calendar seems booked. If not, that's fine to hash out here what this will look like:

(1) So, what I think you're saying is to transfer the whole LJ example to the `hydragnn/examples` directory, including dataset creation, train.py, and json files. This will pull from both your fork and `PyTorch_interatomic_potentials`. It should be able to run similar to the qm9 example,  where `____.py` will detect if there is data and generate it if not. 
(2) Then, would the `examples/LJ` dataset be simply copy-pasted over to `hydragnn/serialized_dataset` to use for examples, or is that unrelated? It would be 500 data points in line with the other examples

hydragnn/models/DIMEStack.py
```diff
@@ -101,7 +101,7 @@ def get_conv(self, input_dim, output_dim):
             out_emb_channels=self.out_emb_size,
             out_channels=output_dim,
             num_layers=1,
-            act=SiLU(),
+            act=Sigmoid(),  # Sigmoid instead of SiLU here promotes stability when we have a linear decoder at the start of convolution (especially for random data examples in test_graphs.py)
```
allaffa, line 104: Is this really necessary?

examples/LennardJones/Forces_Scatterplot.png
allaffa, line 1: @RylieWeaver 
No pictures

examples/LennardJones/LJ_vlad_atomic_forces.json
allaffa, line 1: @RylieWeaver 
Do you need this file?

examples/LennardJones/LJ_multitask.json
allaffa, line 1: @RylieWeaver 
Do you need this file?

examples/LennardJones/LJ_vlad_total_energy.json
allaffa, line 1: @RylieWeaver 
Do you need this file?

examples/LennardJones/train_vlad_atomic_forces.py
allaffa, line 1: @RylieWeaver 
Do you need this file?

examples/LennardJones/train_vlad_total_energy.py
allaffa, line 1: @RylieWeaver 
Do you need this file?

allaffa (changes requested):

examples/LennardJones/train_vlad_total_energy.py
RylieWeaver: @allaffa No, there are a lot of things that I don't think are necessary here, but I wanted to ask before taking out. Does it sound good to take out all the `Vlad` files?

examples/LennardJones/graph_utils.py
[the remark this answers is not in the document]
RylieWeaver: @allaffa Thanks, I'll look into that.

examples/LennardJones/Forces_Scatterplot.png
RylieWeaver: @allaffa sounds good

examples/LennardJones/train_vlad_total_energy.py
RylieWeaver: @allaffa Additionally, I think we could take out `inference.py`, since it would be a rare use case to only want to see predicted energy but not forces.

RylieWeaver:
@allaffa I seem to be getting an error relating to dataset creation on GitHub. Have you ever seen something like this before?

examples/LennardJones/distributed_utils.py
```diff
@@ -0,0 +1,3 @@
+def nsplit(a, n):
```
allaffa, line 1: @RylieWeaver 
This function is already included inside "hydragnn/utils/distributed.py"
Import it from there without re-implementing it.

examples/LennardJones/graph_utils.py
```diff
@@ -0,0 +1,56 @@
+import torch
```
allaffa, line 1: @RylieWeaver 
These functions are already implemented in `hydragnn/preprocess/utils.py`
Can you use those implementations or did you make any changes?

allaffa (changes requested):
@RylieWeaver 
Additional changes that you forgot to address in the preview review of the PR

examples/LennardJones/LJ_inference_plots.py
```diff
@@ -0,0 +1,243 @@
+##############################################################################
+# Copyright (c) 2021, Oak Ridge National Laboratory                          #
```
allaffa, line 2:
```suggestion
# Copyright (c) 2024, Oak Ridge National Laboratory                          #
```

allaffa (approved):
@RylieWeaver 
I committed just one tiny change on the year of the Copyright.

RylieWeaver:
> @RylieWeaver I committed just one tiny change on the year of the Copyright.

@allaffa Ok, cool, I just did a couple of cleanup things as well but it should pass all tests again.

allaffa (approved):

allaffa:
@pzhanggit @jychoi-hpc 
I have been working with Rylie on this PR for quite some time. 
Even if nominally all commits are authored by him, some portions of code were written by me in local script that I passed to him. 
Because of this, sine I have been heavily involved in the code development of this PR, I would appreciate the opinion of fresh eyes, since you may catch issues or concerns in terms of software design that I have not caught myself.

RylieWeaver:
Double-clicking on this, let me know what you think @pzhanggit @jychoi-hpc 

There are some more PRs in waiting that require this one first, so I'd like to get this one approved or revised soon if possible. Thanks.

allaffa:
> Double-clicking on this, let me know what you think @pzhanggit @jychoi-hpc
> 
> There are some more PRs in waiting that require this one first, so I'd like to get this one approved or revised soon if possible. Thanks.

@pzhanggit and @jychoi-hpc 

For the interest of time, please focus only on reviewing only the files that change the core capabilities of hydragnn:

- `hydragnn/models/Base.py`
- `hydragnn/train/train_validate_test.py`

All the other files are in the "examples" folder, therefore they do no affect the core of HydraGNN. The reason why I asked for you to review the PR is because of the two files above.

jychoi-hpc (approved):
Thank you for the update. I looked over the base and train script. It looks good to me.

hydragnn/models/Base.py
```diff
@@ -355,6 +356,60 @@ def loss(self, pred, value, head_index):
...
+        tasks_loss.append(self.loss_function(graph_energy_pred, graph_energy_true))
+        # Forces
+        forces_true = data.forces.float()
+        forces_pred = torch.autograd.grad(
+            graph_energy_pred,
+            data.pos,
+            grad_outputs=torch.ones_like(graph_energy_pred),
+            retain_graph=graph_energy_pred.requires_grad,  # Retain graph only if needed (it will be needed during training, but not during validation/testing)
+            create_graph=True,
+        )[0].float()
```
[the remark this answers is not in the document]
RylieWeaver: Here's my understanding of it: 
The `grad_outputs` argument essentially serves as a weight to multiply the gradients. This is necessary since `graph_energy_pred` is a higher dimension than just a scalar. The _torch.ones_like()_ here will assign a weight of 1 to each gradient, which I believe is what we want. Otherwise, there would be a force prediction which is c*grad_E where c is a constant not equal to 1, which is nonphysical.

Specifically to your question:
No, I don't believe that it results in a sum. The output shape of forces_pred will be in the same shape as data.pos. The grad_outputs specifically is for weighting the gradients before a sum, but that extra step of summing is outside of autograd calculation I believe.

Does this answer your question?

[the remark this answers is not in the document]
pzhanggit: > I don't think so.
> 
> Here's my understanding of it: The `grad_outputs` argument essentially serves as a weight to multiply the gradients. This is necessary since `graph_energy_pred` is a higher dimension than just a scalar. The _torch.ones_like()_ here will assign a weight of 1 to each gradient, which I believe is what we want. Otherwise, there would be a force prediction which is c*grad_E where c is a constant not equal to 1, which is nonphysical.
> 
> Specifically to your question: No, I don't believe that it results in a sum. The output shape of forces_pred will be in the same shape as data.pos. The grad_outputs specifically is for weighting the gradients before a sum, but that extra step of summing is outside of autograd calculation I believe.
> 
> Does this answer your question?
> 
> @pzhanggit

Thanks. So just to verify if my understanding is correct: `grad_outputs` is needed because we're calculating gradients of a batch of samples here. And since `grad_outputs should be a sequence of length matching output containing the “vector” in vector-Jacobian product`, it provides a way to aggregate/sum the gradients across all the samples in the batch. This is equivalent to calculate the gradients iteratively for each sample, since the cross-gradients between samples would be zero.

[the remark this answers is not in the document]
RylieWeaver: @pzhanggit 

>> grad_outputs is needed because we're calculating gradients of a batch of samples here
Yep, that's right.

>> And since grad_outputs should be a sequence of length matching output containing the “vector” in vector-Jacobian product
I think so. Yes, `grad_outputs` should be a sequence, and it should be the same shape as whatever you're predicting. It then multiplies the matrix of gradients (row-wise, not normal matrix multiplication), which I think is the `vector-Jacobian product` you're referring to.

>> it provides a way to aggregate/sum the gradients across all the samples in the batch. 
I think yes. It's scaling those gradients, which would be relevant in an aggregation/sum. Although, it does not do that aggregation/sum.

>> This is equivalent to calculate the gradients iteratively for each sample, since the cross-gradients between samples would be zero.
Yep.

Are there any parts still unclear?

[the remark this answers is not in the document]
pzhanggit: > @pzhanggit
> 
> > > grad_outputs is needed because we're calculating gradients of a batch of samples here
> 
> Yep, that's right.
> 
> > > And since grad_outputs should be a sequence of length matching output containing the “vector” in vector-Jacobian product
> 
> I think so. Yes, `grad_outputs` should be a sequence, and it should be the same shape as whatever you're predicting. It then multiplies the matrix of gradients (row-wise, not normal matrix multiplication), which I think is the `vector-Jacobian product` you're referring to.
> 
> > > it provides a way to aggregate/sum the gradients across all the samples in the batch.
> 
> I think yes. It's scaling those gradients, which would be relevant in an aggregation/sum. Although, it does not do that aggregation/sum.
> 
> > > This is equivalent to calculate the gradients iteratively for each sample, since the cross-gradients between samples would be zero.
> 
> Yep.
> 
> Are there any parts still unclear?

`Scaling those gradients`?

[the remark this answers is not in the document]
allaffa: The scaling is done with `torch.ones_like(graph_energy_pred)`
In our case, we use vectors of ones because we do NOT want to scale. However, you could apply any multiplying factor (or even provide a customized vector with different values for each entry).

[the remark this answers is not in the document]
RylieWeaver: @pzhanggit @allaffa ^ on what Max said. Also, feel free to @ me so In reply faster :)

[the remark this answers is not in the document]
RylieWeaver: Yep. Wow, the o1-preview is great

pzhanggit (approved):

hydragnn/models/Base.py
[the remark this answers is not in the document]
pzhanggit: thanks.

PR #282, opened 2024-09-20
Forces Tests

RylieWeaver:
This conducts force prediction tests. It directly pulls from the Lennard Jones example, rather than creating its own data independently as done in test_graphs.py. The model stacks that work with this are SchNet, EGNN, DimeNet, and PNAPlus.


RylieWeaver:
Mistaken PR

PR #295, opened 2024-10-21
omat24 example

allaffa:
@ArCho48 @RylieWeaver 
I would like for you to take a look at how the code is structured, so that you can start getting acquainted with how we import new datasets in HydraGNN.

### DIFF
--- a/examples/omat24/train.py
+++ b/examples/omat24/train.py
@@ -163,18 +163,6 @@ if __name__ == "__main__":
         "--inputfile", help="input file", type=str, default="omat24_energy.json"
     )
     parser.add_argument(
-        "--train_path",
-        help="path to training data",
-        type=str,
-        default="s2ef_train_200K_uncompressed",
-    )
-    parser.add_argument(
-        "--test_path",
-        help="path to testing data",
-        type=str,
-        default="s2ef_val_id_uncompressed",
-    )
-    parser.add_argument(
         "--energy_per_atom",
         help="option to normalize energy by number of atoms",
         type=bool,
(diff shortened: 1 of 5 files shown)


RylieWeaver:
@allaffa

Thank you for pointing it out. I will put down my main findings to clarify if I'm understanding things correctly.

Overall, we now support download links and use a FAIRChem class to parse the molecular objects in new datasets.
(1) Download links are defined inside lists and downloaded in a loop based on the task (train/val)
(2) We use the `AseDBDataset`, which gives us functions to easily get traits such as atomic number, position, energy, forces, etc...

examples/omat24/train.py
```diff
@@ -0,0 +1,452 @@
...
+                )
+            )
+
+            rx = list(nsplit(list(range(dataset.num_samples)), self.world_size))[
+                self.rank
+            ]
+
+            for index in iterate_tqdm(rx, verbosity_level=2):
+                try:
+                    xyz = torch.tensor(dataset.get_atoms(index).get_positions())
```
jychoi-hpc, line 115: Can you add "dtype=torch.float32" here? 
I got an error: "RuntimeError: mat1 and mat2 must have the same dtype, but got Double and Float"

allaffa:
```suggestion
                    xyz = torch.tensor(dataset.get_atoms(index).get_positions(), dtype=torch.float32)
```

jychoi-hpc (approved):
I just finished testing. I was able to generate the adios file and run training. The code looks good. Thank you for the update.

PR #297, opened 2024-10-29
Fix pbc conv args

RylieWeaver:
This PR aims to address the issue related to generating conv_args that are in-line with periodic boundary conditions. 

The following implementation should be robust under the following assumptions, which I think should always be true:
(1) radius < min(supercell_size) / 2
(2) The edge_index has been generated correctly with periodic boundary conditions.

Currently, only two models are updated to use the new functionality. If this is a good solution, it can be easily extended to the others.


RylieWeaver:
@allaffa This should give an idea of the implementation I've been working with. I can still take a further look at the Facebook implementation you mentioned, but since this one is already working and I'm familiar with I wanted to show you what it would look like first. Let me know your thoughts.

Also I am not sure why there are so many commits. Maybe I forgot a rebase or merge

PR #302, opened 2024-10-29
Light refactor to enhance development

JustinBakerMath:
I'm initializing a PR with the following initiatives:

 (1) light refactoring as discussed this morning (args -> create.py)
 (2) equivariance clean up (inv_feat + geom_feat tracking inside of the MPNN)

I will complete this PR and add full documentation if there is interest in these initiatives.

### DIFF
--- a/hydragnn/models/PAINNStack.py
+++ b/hydragnn/models/PAINNStack.py
@@ -44,7 +44,9 @@ class PAINNStack(Base):
 
     def _init_conv(self):
         last_layer = 1 == self.num_conv_layers
-        self.graph_convs.append(self.get_conv(self.input_dim, self.hidden_dim))
+        self.graph_convs.append(
+            self.get_conv(self.input_dim, self.hidden_dim, last_layer)
+        )
         self.feature_layers.append(nn.Identity())
         for i in range(self.num_conv_layers - 1):
             last_layer = i == self.num_conv_layers - 2


hydragnn/models/EGCLStack.py
```diff
@@ -30,7 +32,11 @@ def __init__(
         self.edge_dim = (
             0 if edge_attr_dim is None else edge_attr_dim
         )  # Must be named edge_dim to trigger use by Base
-        super().__init__(*args, **kwargs)
+        super().__init__(input_args, conv_args, *args, **kwargs)
+
+        assert (
+            self.input_args == "inv_node_feat, equiv_node_feat, edge_index, edge_attr"
```
RylieWeaver, line 38: @JustinBakerMath 

Does EGCL always require edge_attributes?

hydragnn/models/PAINNStack.py
```diff
@@ -75,82 +77,54 @@ def get_conv(self, input_dim, output_dim, last_layer=False):
...
+                        lambda inv_node_feat, equiv_node_feat: [
+                            inv_node_feat,
+                            equiv_node_feat,
+                        ],
+                        "inv_node_feat, equiv_node_feat -> inv_node_feat, equiv_node_feat",
+                    ),
                 ],
             )
 
-    def forward(self, data):
```
RylieWeaver, line 104: @JustinBakerMath 

This is the largest advantage of the approach you're taking here, right? By generalizing pos/v --> equiv_node_feature, you've taken out the need to rewrite the whole forward function.l

RylieWeaver:
@JustinBakerMath 

Are the model stacks like PAINN, DimeNet, etc... finished with conversion to your proposed method here? If so, then I can take inspiration and adjust MACE tomorrow :)

JustinBakerMath:
Yes. I will change the extra formatting back. I ran into some other issues with PAINN and PNAEq as observed in the most recent check. I ran out of time to reason through it yesterday. I will check it out this evening.

I have temporarily converted MACE to this format but I believe there is a better way to handle the variables. I would definitely appreciate help cleaning up MACE. Let me ping you when the checks pass for PAINN and PNAEq.

hydragnn/models/PAINNStack.py
JustinBakerMath: The aim is to support the development equivariant architectures by providing a specific variable for handling equivariant information. This is done by improving the variable names to `inv_node_feat` and `equiv_node_feat` instead of vaguely named `x, pos`, respectively. One advantage would be that for many architectures, the forward function is handled by the Base class.

PR #306, opened 2024-11-06
Fix pbc conv args Final Draft

RylieWeaver:
Fix pbc conv args final draft

Main points:

### (1) I create edge_shifts in preprocessing with ASE, then apply those shifts in conv_args.
 ------The edge shifts adjust the relative vector for a connection. After additional research, I believe this to be the best way for fewer lines of code and being streamlined. It is also the approach of top competitors and does not disrupt gradient flow------

### (2) Correctly applying PBC in models that update the position is currently not supported (SchNet / EGNN). 
  ------At the core level, we have the problem that the positional update could make a node cross the supercell, even if we were to recalculate the shifts after each layer. I still allow the models to run, but simply continue to use the naive relative vectors------

### (3) data.pbc is now a requirement for periodic boundary conditions
 ------I make an effort to include it in any examples that use PBC------

@allaffa Assuming this passes the tests I'll schedule a meeting to go over the PR.

### DIFF
--- a/examples/LennardJones/LJ_data.py
+++ b/examples/LennardJones/LJ_data.py
@@ -173,7 +173,7 @@ class LJDataset(AbstractBaseDataset):
             .unsqueeze(0)
             .to(torch.float32),
             energy=torch.tensor(total_energy).unsqueeze(0).to(torch.float32),
-            pbc=True,
+            pbc=[True, True, True],
         )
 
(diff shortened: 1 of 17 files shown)


allaffa:
@RylieWeaver 
After merging the PR for MACE. conflicts popped up on this PR

allaffa (approved):

PR #310, opened 2024-11-26
edge_dim added as argument to GATv2Conv

allaffa:
This PR applies two major changes: 

- allows to pass `edge_dim` as argument to GATStack because the original PyG implementation allows to use edge attributes [GATv2Conv](https://pytorch-geometric.readthedocs.io/en/2.5.1/generated/torch_geometric.nn.conv.GATv2Conv.html)
- parametrizes the examples `qm9`, `md17,` and `LennardJones` over all the message passing layers

These changes are crucial to support the future integration go graph transformer architectures and allow a robust testing of those capabilities that should flexibly switch across all message passing layers.


hydragnn/utils/input_config_parsing/config_utils.py
ArCho48, line 1: Perhaps we can add GAT in this assert message in line 165: "Edge features can only be used with DimeNet, MACE, EGNN, SchNet, PNA, PNAPlus, PNAEq, and CGCNN."

ArCho48 (approved):

RylieWeaver:
@allaffa 

### The fixes you put in place look good to me. For brevity here's my understanding:

It looks like you identified the following Errors and Fixes:
(1) Some MP-layers may not inherently support the existence of positional gradients. In this case, you exclude them from the LJ-example. This fixed the errors with: PNA, SAGE, GIN, GAT, MFC
(2) The examples did not allow you to choose the MP-layer via a command-line-argument. In this case, you added this optionality. This fixed the errors with PNAPlus, PNAEq



### Additionally, I added the following in my commits:

(1) Fixes to do with **object device placement** during degree aggregation in distributed data processing and message-passing. This was not causing errors in only CPU-environments like these tests, but would cause errors in GPU environments and was identified when testing on the DGX.
(2) Squeezed the energy predictions in Base to to avoid the shape mismatch warning.
(3) Add a comment explaining which types of MP-layers are allowed in the `grad_energy` example in `test_examples.py`



Lastly, I wasn't able to reproduce the following warning to fix. Do we still want to tackle it?:
- `/home/runner/work/HydraGNN/HydraGNN/hydragnn/utils/model/model.py:178: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).`

allaffa:
> @allaffa @ArCho48
> 
> ### Max, the fixes you put in place look good to me. For brevity here's my understanding:
> It looks like you identified the following Errors and Fixes: (1) Some MP-layers may not inherently support the existence of positional gradients. In this case, you exclude them from the LJ-example. This fixed the errors with: PNA, SAGE, GIN, GAT, MFC (2) The examples did not allow you to choose the MP-layer via a command-line-argument. In this case, you added this optionality. This fixed the errors with PNAPlus, PNAEq
> 
> ### Additionally, I added the following in my commits:
> (1) Fixes to do with **object device placement** during degree aggregation in distributed data processing and message-passing. This was not causing errors in only CPU-environments like these tests, but would cause errors in GPU environments and was identified when testing on the DGX. (2) Squeezed the energy predictions in Base to to avoid the shape mismatch warning. (3) Add a comment explaining which types of MP-layers are allowed in the `grad_energy` example in `test_examples.py` (4) Added MACE to the `grad_energy` example because it will propagate a positional gradient.
> 
> Lastly, I wasn't able to reproduce the following warning to fix. Do we still want to tackle it?:
> 
> * `/home/runner/work/HydraGNN/HydraGNN/hydragnn/utils/model/model.py:178: UserWarning: To copy construct from a tensor, it is recommended to use sourceTensor.clone().detach() or sourceTensor.clone().detach().requires_grad_(True), rather than torch.tensor(sourceTensor).`

@RylieWeaver 
Thank you for helping me out in this PR.

ArCho48:
Thank you both!

PR #313, opened 2024-12-03
Align examples pbc

RylieWeaver:
This PR aims to accomplish four main things:

(1) Align the OC examples with their datasets in the literature, which always have PBC
(2) Unify the naming of `cell` in the examples and all other parts of HydraGNN. Including `cell` is required by PBC
(3) Improved simplicity of the LJ example
  -  Using tools for `edge_vec` and `forces`, rather than manual
(4) Improved robustness of applying pbc
  - Checking for dupe edges instead of a check on radius, since the radius check can falter for non-cubic PBC
  - Adjusting the radius in OC examples so as to not create dupe edges

### DIFF
--- a/examples/LennardJones/LJ_data.py
+++ b/examples/LennardJones/LJ_data.py
@@ -186,11 +186,14 @@ class LJDataset(AbstractBaseDataset):
             .unsqueeze(0)
             .to(torch.float32),
             energy=torch.tensor(total_energy).unsqueeze(0).to(torch.float32),
-            pbc=[
-                True,
-                True,
-                True,
-            ],  # LJ example always has periodic boundary conditions
+            pbc=torch.tensor(
+                [
+                    True,
+                    True,
+                    True,
+                ],
+                dtype=torch.bool,
+            ),  # LJ example always has periodic boundary conditions
         )
 
         # Create pbc edges and lengths
@@ -352,30 +355,21 @@ def create_configuration(
     data.cell = torch.diag(
         torch.tensor([supercell_size_x, supercell_size_y, supercell_size_z])
     )
-    data.pbc = [True, True, True]
+    data.pbc = torch.tensor([True, True, True], dtype=torch.bool)
+    data.x = torch.cat([atom_types, positions], dim=1)
 
     create_graph_connectivity_pbc = get_radius_graph_pbc(
         radius_cutoff, max_num_neighbors
     )
     data = create_graph_connectivity_pbc(data)
 
-    atomic_descriptors = torch.cat(
-        (
-            atom_types,
-            positions,
-        ),
-        1,
-    )
-
-    data.x = atomic_descriptors
-
     data = atomic_structure_handler.compute(data)
 
     total_energy = torch.sum(data.x[:, 4])
     energy_per_atom = total_energy / number_nodes
 
-    total_energy_str = numpy.array2string(total_energy.detach().numpy())
-    energy_per_atom_str = numpy.array2string(energy_per_atom.detach().numpy())
+    total_energy_str = numpy.array2string(total_energy.detach().cpu().numpy())
+    energy_per_atom_str = numpy.array2string(energy_per_atom.detach().cpu().numpy())
     filetxt = total_energy_str + "\n" + energy_per_atom_str
 
     for index in range(0, 3):
--- a/examples/LennardJones/LJ_data.py
+++ b/examples/LennardJones/LJ_data.py
@@ -53,6 +53,7 @@ primitive_bravais_lattice_constant_z = 3.8
 
 def create_dataset(path, config):
     radius_cutoff = config["NeuralNetwork"]["Architecture"]["radius"]
+    max_num_neighbors = config["NeuralNetwork"]["Architecture"]["max_neighbours"]
     number_configurations = (
         config["Dataset"]["number_configurations"]
         if "number_configurations" in config["Dataset"]
@@ -75,6 +76,7 @@ def create_dataset(path, config):
         atom_types,
         atomic_structure_handler=atomic_structure_handler,
         radius_cutoff=radius_cutoff,
+        max_num_neighbors=max_num_neighbors,
         relative_maximum_atomic_displacement=1e-1,
         number_configurations=number_configurations,
     )
--- a/examples/LennardJones/LJ_data.py
+++ b/examples/LennardJones/LJ_data.py
@@ -403,7 +403,6 @@ class AtomicStructureHandler:
         self.bravais_lattice_constants = bravais_lattice_constants
         self.radius_cutoff = radius_cutoff
         self.formula = formula
-    
 
     # Calculate the potential energy with torch gradient tracking, then simply use autograd to calculate the forces
     def compute(self, data):
@@ -411,28 +410,38 @@ class AtomicStructureHandler:
         assert data.pos.shape[0] == data.x.shape[0]
         node_potential = torch.zeros([data.pos.shape[0], 1])
         node_forces = torch.zeros([data.pos.shape[0], 3])
-        
+
         # Calculate
         data.pos.requires_grad = True
         edge_vec, edge_dist = get_edge_vectors_and_lengths(
             positions=data.pos,
             edge_index=data.edge_index,
             shifts=data.edge_shifts,
-            normalize=False
+            normalize=False,
         )
-        
+
         # Sum potential by edge, node, and total
-        edge_potential = self.formula.potential_energy(edge_dist)  # Shape [num_edges, 1]
-        node_potential = scatter(edge_potential, data.edge_index[0], dim=0, dim_size=data.pos.shape[0], reduce='add')  # Shape [num_nodes, 1]
+        edge_potential = self.formula.potential_energy(
+            edge_dist
+        )  # Shape [num_edges, 1]
+        node_potential = scatter(
+            edge_potential,
+            data.edge_index[0],
+            dim=0,
+            dim_size=data.pos.shape[0],
+            reduce="add",
+        )  # Shape [num_nodes, 1]
         total_potential = torch.sum(node_potential, dim=0, keepdim=True)  # Shape [1]
-        
+
         # Autograd to calculate forces
         node_forces = -torch.autograd.grad(
             total_potential,
             data.pos,
             grad_outputs=torch.ones_like(total_potential),
-        )[0]  # Shape [num_nodes, 3]
-        
+        )[
+            0
+        ]  # Shape [num_nodes, 3]
+
         # Append to data
         data.x = torch.cat((data.x, node_potential, node_forces), dim=1)
 
@@ -443,14 +452,13 @@ class LJpotential:
     def __init__(self, epsilon, sigma):
         self.epsilon = epsilon
         self.sigma = sigma
-        
+
     def potential_energy(self, pair_distance):
         return (
             4
             * self.epsilon
             * ((self.sigma / pair_distance) ** 12 - (self.sigma / pair_distance) ** 6)
         )
-
 
 
 """Etc"""
--- a/examples/LennardJones/LJ_data.py
+++ b/examples/LennardJones/LJ_data.py
@@ -167,7 +167,7 @@ class LJDataset(AbstractBaseDataset):
         forces_pre_scaled = forces * forces_pre_scaling_factor
 
         data = Data(
-            supercell_size=torch_supercell.to(torch.float32),
+            cell=torch_supercell.to(torch.float32),
             num_nodes=num_nodes,
             grad_energy_post_scaling_factor=grad_energy_post_scaling_factor,
             forces_pre_scaling_factor=torch.tensor(forces_pre_scaling_factor).to(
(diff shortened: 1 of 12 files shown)


allaffa (approved):

RylieWeaver:
@allaffa

(1) I unified the naming to `cell` everywhere I could find by `grep -r ***` any relevant keywords like pbc, periodic, etc... However, it would be good to get your opinion if you think there's anywhere I missed.

(2) I think we can reasonably assume that OC has pbc for every data sample, since I've checked on preprocessing and training `s2ef 200k` and we have quotes like below from the OC papers that indicate so. However, it's not possible to say for sure without a Distributed Data Parallel run on a bigger machine, since the dataset is so large. I just wanted to give a heads-up on this fact and the length of testing that has been done.

OC Quote:
In all our tasks, the structure contains a surface and an adsorbate. The surface is defined by a unit cell that is periodic in all directions with a vacuum layer of at least 20 Å applied in the z-direction. Initial structures are heuristically determined. Ground truth data are computed for all tasks using DFT. Dataset details and evaluation metrics are provided in following sections.
[link](https://pubs.acs.org/doi/10.1021/acscatal.0c04525)

allaffa:
@RylieWeaver 
Please apply the same changes to the following examples:

- alexandria
- omat24
- mptrj

allaffa:
@RylieWeaver 
Thank you. 

I have one more question. What happens if, accidentally, the combination of cell size and radius cut-off induce the cell to be relatively too small in size, such that PBC cannot be applied?

RylieWeaver:
> @RylieWeaver 
> 
> Thank you. 
> 
> 
> 
> I have one more question. What happens if, accidentally, the combination of cell size and radius cut-off induce the cell to be relatively too small in size, such that PBC cannot be applied?

In the case that the cell is relatively too small compared to the radius, there will be duplicate edges and the code will throw an assert error from the check in 'get_edge_vectors_and_lengths'.

allaffa:
> > @RylieWeaver
> > Thank you.
> > I have one more question. What happens if, accidentally, the combination of cell size and radius cut-off induce the cell to be relatively too small in size, such that PBC cannot be applied?
> 
> In the case that the cell is relatively too small compared to the radius, there will be duplicate edges and the code will throw an assert error from the check in 'get_edge_vectors_and_lengths'.

So essentially you are telling me that EVERY example will throw an error, because this case will happen for every dataset that this PR addresses :) This is not very helpful 

For each example, you need to deal with this situation, by switching off the periodic boundary conditions for each data sample when PBC cannot be applied.

RylieWeaver:
> > > @RylieWeaver
> > > Thank you.
> > > I have one more question. What happens if, accidentally, the combination of cell size and radius cut-off induce the cell to be relatively too small in size, such that PBC cannot be applied?
> > 
> > 
> > In the case that the cell is relatively too small compared to the radius, there will be duplicate edges and the code will throw an assert error from the check in 'get_edge_vectors_and_lengths'.
> 
> So essentially you are telling me that EVERY example will throw an error, because this case will happen for every dataset that this PR addresses :) This is not very helpful
> 
> For each example, you need to deal with this situation, by switching off the periodic boundary conditions for each data sample when PBC cannot be applied.

Ok, I will return with a better implementation by tonight. I cannot speak to every example, but confirmed that there were a range of radii which worked for every sample in OC2020 200k. 

Still, I think what I am understanding is that it's better to have a robust example which allows for some incorrectness than a brittle one which requires more careful knowledge by the user?

Thank you for the feedback, it is good to fix now. I will get back soon with something satisfactory carrying out this philosophy :)

allaffa:
> > > > @RylieWeaver
> > > > Thank you.
> > > > I have one more question. What happens if, accidentally, the combination of cell size and radius cut-off induce the cell to be relatively too small in size, such that PBC cannot be applied?
> > > 
> > > 
> > > In the case that the cell is relatively too small compared to the radius, there will be duplicate edges and the code will throw an assert error from the check in 'get_edge_vectors_and_lengths'.
> > 
> > 
> > So essentially you are telling me that EVERY example will throw an error, because this case will happen for every dataset that this PR addresses :) This is not very helpful
> > For each example, you need to deal with this situation, by switching off the periodic boundary conditions for each data sample when PBC cannot be applied.
> 
> Ok, I will return with a better implementation by tonight. I cannot speak to every example, but confirmed that there were a range of radii which worked for every sample in OC2020 200k.
> 
> Still, I think what I am understanding is that it's better to have a robust example which allows for some incorrectness than a brittle one which requires more careful knowledge by the user?
> 
> Thank you for the feedback, it is good to fix now. I will get back soon with something satisfactory carrying out this philosophy :)

In each example, there are three scenarios that need to be dealt with:
1. We deliberately do NOT intend to apply PBC --> generate the torch geometric data WITHOUT applying the PBC
2. We intend to apply PBC and the cell size is large enough to allow for it --> generate the torch geometric data WITH PBC
3. We  intend to apply PBC and BUT cell size of a data sample is TOO SMALL --> generate the torch geometric data WITHOUT applying the PBC, and  print out a message to communicate this

RylieWeaver:
> > > > > @RylieWeaver
> 
> > > > > Thank you.
> 
> > > > > I have one more question. What happens if, accidentally, the combination of cell size and radius cut-off induce the cell to be relatively too small in size, such that PBC cannot be applied?
> 
> > > > 
> 
> > > > 
> 
> > > > In the case that the cell is relatively too small compared to the radius, there will be duplicate edges and the code will throw an assert error from the check in 'get_edge_vectors_and_lengths'.
> 
> > > 
> 
> > > 
> 
> > > So essentially you are telling me that EVERY example will throw an error, because this case will happen for every dataset that this PR addresses :) This is not very helpful
> 
> > > For each example, you need to deal with this situation, by switching off the periodic boundary conditions for each data sample when PBC cannot be applied.
> 
> > 
> 
> > Ok, I will return with a better implementation by tonight. I cannot speak to every example, but confirmed that there were a range of radii which worked for every sample in OC2020 200k.
> 
> > 
> 
> > Still, I think what I am understanding is that it's better to have a robust example which allows for some incorrectness than a brittle one which requires more careful knowledge by the user?
> 
> > 
> 
> > Thank you for the feedback, it is good to fix now. I will get back soon with something satisfactory carrying out this philosophy :)
> 
> 
> 
> In each example, there are three scenarios that need to be dealt with:
> 
> 1. We deliberately do NOT intend to apply PBC --> generate the torch geometric data WITHOUT applying the PBC
> 
> 2. We intend to apply PBC and the cell size is large enough to allow for it --> generate the torch geometric data WITH PBC
> 
> 3. We  intend to apply PBC and BUT cell size of a data sample is TOO SMALL --> generate the torch geometric data WITHOUT applying the PBC, and  print out a message to communicate this

If there is time, let's connect on this at the end of our meeting tomorrow.

On the 3:

(1) I thought that the 'periodic_boundary_conditions' parameter was deprecated, and that PBC should be made to align with the example without optionality? This way, a user could not make a nonsensical parameter choice given the data.
(2) Yep.
(3) Either this or we allow for dupe edges. After further research, I believe our original assumption that dupe edges are not allowed is incorrect, as the top competitors Nequip, MACE, matsciml, don't see to do so.

examples/mptrj/train.py
```diff
@@ -121,9 +122,20 @@ def __init__(
...
-                lattice_mat = torch.tensor(
-                    info["atoms"]["lattice_mat"], dtype=torch.float32
-                )
+                lattice_mat = None
+                try:
+                    lattice_mat = torch.tensor(
+                        info["atoms"]["lattice_mat"], dtype=torch.float32
+                    )
+                except:
+                    print(f"Structure does not have lattice_mat", flush=True)
```
RylieWeaver, line 131: @allaffa 

I took inspiration from your Alexandria implementation here. Do you like `flush=True` for all of these examples?

hydragnn/preprocess/graph_samples_checks_and_updates.py
```diff
@@ -171,29 +166,64 @@ def __call__(self, data):
...
 
-        # ensure no duplicate edges
-        unique_edge_index, unique_indices = torch.unique(
-            data.edge_index, dim=1, return_inverse=False  # Shape: [n_edges]
+        # Assign to data
+        data.edge_index = torch.stack(
+            [edge_src, edge_dst],
+            dim=0,  # Shape: [2, n_edges]
         )
-        assert unique_edge_index.unsqueeze(0).size(1) == data.edge_index.size(
```
RylieWeaver, line 183: @allaffa 

The assert has been removed so that pbc is robustly applied to examples that expect it without restrictions on the radius.

hydragnn/preprocess/graph_samples_checks_and_updates.py
```diff
@@ -171,29 +166,64 @@ def __call__(self, data):
...
+        # ASE returns the integer number of cell shifts. Multiply by the cell size to get the shift vector.
         data.edge_shifts = torch.matmul(
-            torch.tensor(edge_cell_shifts).float(), data.supercell_size.float()
+            edge_cell_shifts, data.cell.float()
         )  # Shape: [n_edges, 3]
 
         return data
 
+    # Function to keep only the closest `max_num_neighbors` per node
+    def _limit_neighbors(
```
RylieWeaver, line 195: @allaffa

This function is essential in providing robustness to increasing radius sizes. When radius graphs are built with pbc at increasing radiuses, the number of neighbors may become computationally very expensive and time-consuming. By maximizing the number of neighbors, we cap the computational and time cost, even if the user defines very large radii relative to the cell size such as in HPO. This increases robustness and, since, we retain the closest neighbors, does not distort the metric for establishing connectivity.

examples/mptrj/train.py
allaffa: Yes. `flush=True` helps makes sure that printouts are not delayed too much when the code runs in parallel at extreme scale.

hydragnn/preprocess/graph_samples_checks_and_updates.py
allaffa: @RylieWeaver 
Does it mean that we run the risk of duplicated edges?

RylieWeaver: @allaffa

Yes. I remember a conservation we had where we talked about this as undesirable, but I believe now this is the functionality we want. I was curious why other models did not have such error handling if there is such a  restrictive assert, and to my surprise, they did not have any such a restriction.

Fundamentally, it seems fine to connect to the 'same node' multiple times in different cells, with different distances and relative vectors for the edges. 

Here's the code that I'm referencing:
[MACE](https://github.com/ACEsuit/mace/blob/main/mace/data/neighborhood.py)
[Nequip](https://github.com/mir-group/nequip/blob/main/nequip/data/AtomicData.py) `def from_points()` (line 325) and `def from_ase()` (line 390)
[MatSciML](https://github.com/IntelLabs/matsciml/blob/main/matsciml/preprocessing/atoms_to_graphs.py) `def _get_neighbors_pymatgen()` (line 83)

examples/LennardJones/LJ_data.py
```diff
@@ -402,73 +400,47 @@ def __init__(
...
+            edge_potential,
+            data.edge_index[0],
+            dim=0,
+            dim_size=data.pos.shape[0],
+            reduce="add",
+        )  # Shape [num_nodes, 1]
+        total_potential = torch.sum(node_potential, dim=0, keepdim=True)  # Shape [1]
+
+        # Autograd to calculate forces
+        node_forces = -torch.autograd.grad(
```
RylieWeaver, line 433: @allaffa 

Instead of manually implementing the equation for force for each node, we can just call autograd. This is easier and reduces the number of lines of code, helping to prevent errors.

examples/open_catalyst_2020/utils/atoms_to_graphs.py
```diff
@@ -51,21 +51,19 @@ class AtomsToGraphs:
     def __init__(
         self,
         max_neigh=200,
-        radius=6,
-        r_pbc=False,
+        radius=6.0,
     ):
         self.max_neigh = max_neigh
         self.radius = radius
-        self.r_pbc = r_pbc
```
RylieWeaver, line 59: @allaffa 

`r_pbc` is no longer needed because our default behavior is to use PBC. This should be the case because Open Catalyst makes all its samples with PBC.

hydragnn/preprocess/graph_samples_checks_and_updates.py
```diff
@@ -170,31 +161,122 @@ def __call__(self, data):
...
+            "ijdS",
+            a=ase_atom_object,
+            cutoff=self.r,
+            self_interaction=True,  # We want self-interactions across periodic boundaries
         )
 
-        # ensure no duplicate edges
-        unique_edge_index, unique_indices = torch.unique(
-            data.edge_index, dim=1, return_inverse=False  # Shape: [n_edges]
+        # Eliminate true self-loops if desired
```
RylieWeaver, line 170: @allaffa

We would like to have self loops from a node to itself in different Periodic Boundary cells, so we make the radius graph with self loops and then remove "true" self loops as desired.

hydragnn/preprocess/graph_samples_checks_and_updates.py
```diff
@@ -170,31 +161,122 @@ def __call__(self, data):
...
+
+        # Apply the mask and return
+        return (
+            edge_src[mask],
+            edge_dst[mask],
+            edge_length[mask],
+            edge_cell_shifts[mask],
+        )
+
+    def _limit_neighbors(
```
RylieWeaver, line 223: @allaffa

We limit the neighbors of nodes manually, since ASE neighbor list will not do so. For each node, we keep only the closest `max_num_neighbors`.

allaffa (approved):

allaffa (approved):

PR #315, opened 2024-12-05
Implementing graphGPS as a global attention wrapper around Hydra MPNNs

ArCho48:
This PR incorporates graphGPS within the HydraGNN framework.


allaffa (approved):

tests/test_graphs.py
```diff
@@ -215,27 +217,48 @@ def unittest_train_model(
...
-def pytest_train_model(model_type, ci_input, overwrite_data=False):
-    unittest_train_model(model_type, ci_input, False, overwrite_data)
+def pytest_train_model(mpnn_type, ci_input, overwrite_data=False):
+    unittest_train_model(mpnn_type, None, None, ci_input, False, overwrite_data)
+
+
+# Test only models
+@pytest.mark.parametrize(
+    "global_attn_engine",
+    ["GPS"],
```
allaffa, line 227: @ArCho48 
We need to try also "None" as global attention mechanism

allaffa (commented):
@ArCho48 
The tests need to check the MPNNs both with and without global attention mechanisms.

tests/test_graphs.py
ArCho48: Screenshot 2024-12-05 at 2 28 47 PM

I thought these lines do that

allaffa: You are right. :) My apologies

ArCho48: Cool. No probs :)

examples/LennardJones/LJ_data.py
```diff
@@ -101,11 +102,19 @@ def __init__(self, dirpath, config, dist=False, sampling=None):
 
         rx = list(nsplit((dirfiles), self.world_size))[self.rank]
 
+        # LPE
+        self.transform = AddLaplacianEigenvectorPE(
```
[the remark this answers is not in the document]
ArCho48: Indeed, lapPE is a purely topological encoding

hydragnn/models/Base.py
```diff
@@ -147,7 +223,24 @@ def _embedding(self, data):
                 data.edge_attr is not None
             ), "Data must have edge attributes if use_edge_attributes is set."
             conv_args.update({"edge_attr": data.edge_attr})
-        return data.x, data.pos, conv_args
+
+        if self.use_global_attn:
+            # encode node positional embeddings
+            x = self.pos_emb(data.pe)
+            # if node features are available, genrate mebeddings, concatenate with positional embeddings and map to hidden dim
+            if self.input_dim:
```
RylieWeaver, line 231: @ArCho48 

Do we expect `input_dim` to ever not be defined? Is this a new type of functionality in Hydra where there are solely graph features and no node features?

examples/LennardJones/LJ_data.py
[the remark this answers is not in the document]
ArCho48: Bdw, @RylieWeaver we couldnt get GPS to run on LennardJones as pytorch throws an error when backpropagating manually computed gradients (which I think is done for LJ) through the attention layers

hydragnn/models/Base.py
ArCho48: graph features or even edge features but no node features. For ex, qm9 doesnt have any node feats as far as I know

hydragnn/models/CGCNNStack.py
```diff
@@ -53,10 +58,12 @@ def __init__(
             assert self.input_args == "inv_node_feat, equiv_node_feat, edge_index"
             assert self.conv_args == "inv_node_feat, edge_index"
 
-    def get_conv(self, input_dim, _):
+    def get_conv(self, input_dim, _, edge_dim=None):
+        if not edge_dim:
```
RylieWeaver, line 62: @ArCho48 

I am not sure, but it may be better to instead handle this inside `__init__()`, by adding a line in the section `if self.global_attn_engine:` that sets `self.edge_dim` with:

```
self.another_dim = self.embed_dim = self.edge_embed_dim = hidden_dim  # ensure that all input to gps have the same dimensionality
```

or something like it. This would get rid of the if/else in every stack's `get_conv` function.

ArCho48: I cannot change self.edge_dim between lines 134-152 of base bcoz I need the original edge_dim in line 167. But I think I can get rid of the if/else in get_conv since edge_dim is always passed to get_conv through the edge_embed_dim variable. Thanks for the suggestion

examples/LennardJones/LJ_data.py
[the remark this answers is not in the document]
RylieWeaver: > Bdw, @RylieWeaver we couldnt get GPS to run on LennardJones as pytorch throws an error when backpropagating manually computed gradients (which I think is done for LJ) through the attention layers

Yes, LJ uses `compute_grad_energy` to compute forces by the gradient of energy wrt position. It is also a general functionality which could be used for other force/energy prediction examples.

As a first guess, I believe the error comes from a lack of gradient flow to the position from the Laplacian Eigenvectors. 

If this is the case, a potentially easy fix might be to do the Laplacian computation with `torch.no_grad()`, like this:
```
with torch.no_grad():
    data = self.transform(data)
```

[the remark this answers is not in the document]
ArCho48: Well, that can be a problem perhaps, but the error is mainly due to the incompatibility of gradient flow with the torch multiheadattention layer. I will send u the exact error shortly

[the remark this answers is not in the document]
RylieWeaver: Hmm, yes, please let me know what comes up.

hydragnn/models/CGCNNStack.py
RylieWeaver: Hmm, good point

allaffa (approved):

examples/LennardJones/LJ_data.py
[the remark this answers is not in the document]
RylieWeaver: @ArCho48 Is this error still coming up? Have you had the opportunity to make the Laplacian with torch.nograd()?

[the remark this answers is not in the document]
ArCho48: Hi Rylie, I think we decided to skip integrating LJ with GPS in this PR. I can work on this (n add ur fix if i works) in the next PR on gps. @allaffa do u have any suggestion?

RylieWeaver (approved):

PR #323, opened 2025-02-03
Dev hierarchical multibranch output heads 

pzhanggit:
to support HydraGNN cases where users need dataset-specific heads for the same variable across multiple datasets.

### DIFF
--- a/examples/multibranch/train.py
+++ b/examples/multibranch/train.py
@@ -1,0 +1,385 @@
+import os, json
+import logging
+import sys
+from mpi4py import MPI
+import argparse
+
+import torch
+
+# FIX random seed
+random_state = 0
+torch.manual_seed(random_state)
+
+import numpy as np
+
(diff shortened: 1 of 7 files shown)


examples/multibranch/train.py
```diff
@@ -0,0 +1,409 @@
...
+        mycolor = colorlist[rank]
+        mymodel = modellist[mycolor]
+
+        local_comm = comm.Split(mycolor, rank)
+        local_comm_rank = local_comm.Get_rank()
+        local_comm_size = local_comm.Get_size()
+
+        ## FIXME: Hard-coded for now. Need to find common variable names
+        common_variable_names = [
+            "x",
```
allaffa, line 237: @pzhanggit @jychoi-hpc 
shouldn't we check also on the presence of `data.energy` and `data.forces` ?

allaffa (approved):

PR #324, opened 2025-02-10
normalization of chemical composition histogram removed from calculation of linear regression term

allaffa:
normalization of chemical composition histogram removed from calculation of linear regression term

### DIFF
--- a/examples/multibranch/energy_linear_regression.py
+++ b/examples/multibranch/energy_linear_regression.py
@@ -1,0 +1,175 @@
+import os, json
+import sys
+from mpi4py import MPI
+import argparse
+
+import torch
+import numpy as np
+
+import hydragnn
+from hydragnn.utils.distributed import nsplit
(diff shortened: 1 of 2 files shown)


pzhanggit (commented):
Thanks @allaffa . The changes look good to me. Do we want to move the `energy_linear_regression.py`  script to the data process part that was used to generate the adios data files, as the script should be applied to all examples and we don't want to keep two versions of energy around?

allaffa:
> Thanks @allaffa . The changes look good to me. Do we want to move the `energy_linear_regression.py` script to the data process part that was used to generate the adios data files, as the script should be applied to all examples and we don't want to keep two versions of energy around?

@pzhanggit 
I prefer to have two sets of ADIOS files. One with the original values, one with the values of the energy change. This allows to treat the linear regression term as an independent module.

pzhanggit (approved):

PR #328, opened 2025-02-13
Fix Nodal Output Dim for Compute-Grad-Energy

RylieWeaver:
This is a draft for how we may fix the nodal MLP's output dimension when using compute_grad_energy. This way, the output dimension is dictated by the config's instead of the node feature dim. We are still setting up dummy data.y and data.y_loc variables, but not using them, but this fix may be minimally invasive.

NOTE: In detail: When using Hydra for physics-informed force prediction for the GFM, we have the following structure:
    # --> Load with ADIOS  -->
    # --> update_predicted_values(): defines y_loc = [0, node_feature_dims[output_index]*num_nodes] -->
    # --> update_config_NN_outputs(): y_loc exists, so it defines output_dim = [(node_feature_dims[output_index]*num_nodes)/num_nodes]  = [node_feature_dims[output_index]] -->
    # --> Base() ... MLPNode(): defines node MLP head with output_dim ... This must be equal to 1 as expected for nodal energy predictions, but is actually equal to the nodal dim at variable of interest because we loaded with ADIOS.

### DIFF
--- a/hydragnn/utils/input_config_parsing/config_utils.py
+++ b/hydragnn/utils/input_config_parsing/config_utils.py
@@ -183,7 +183,10 @@ def update_config_edge_dim(config):
 def check_output_dim_consistent(data, config):
     output_type = config["NeuralNetwork"]["Variables_of_interest"]["type"]
     output_index = config["NeuralNetwork"]["Variables_of_interest"]["output_index"]
-    if hasattr(data, "y_loc"):
+    if (
+        hasattr(data, "y_loc")
+        and not config["NeuralNetwork"]["Training"]["compute_grad_energy"]
+    ):
         for ihead in range(len(output_type)):
             if output_type[ihead] == "graph":
                 assert (
--- a/hydragnn/utils/input_config_parsing/config_utils.py
+++ b/hydragnn/utils/input_config_parsing/config_utils.py
@@ -202,7 +202,9 @@ def update_config_NN_outputs(config, data, graph_size_variable):
     """ "Extract architecture output dimensions and set node-level prediction architecture"""
 
     output_type = config["Variables_of_interest"]["type"]
-    if hasattr(data, "y_loc"):
+    if config["Training"]["compute_grad_energy"]:
+        dims_list = config["Variables_of_interest"]["output_dim"]
+    elif hasattr(data, "y_loc"):
         dims_list = []
         for ihead in range(len(output_type)):
             if output_type[ihead] == "graph":
--- a/hydragnn/utils/input_config_parsing/config_utils.py
+++ b/hydragnn/utils/input_config_parsing/config_utils.py
@@ -183,10 +183,7 @@ def update_config_edge_dim(config):
 def check_output_dim_consistent(data, config):
     output_type = config["NeuralNetwork"]["Variables_of_interest"]["type"]
     output_index = config["NeuralNetwork"]["Variables_of_interest"]["output_index"]
-    if (
-        hasattr(data, "y_loc")
-        and not config["NeuralNetwork"]["Training"]["compute_grad_energy"]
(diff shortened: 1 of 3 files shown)


pzhanggit:
> @pzhanggit
> 
> I've tagged you since we talked at length about this problem previously. Do you mind taking a look at this proposed change and pointing out any potential problems?

Sure, let me know when ready

hydragnn/utils/input_config_parsing/config_utils.py
```diff
@@ -202,7 +202,9 @@ def update_config_NN_outputs(config, data, graph_size_variable):
     """ "Extract architecture output dimensions and set node-level prediction architecture"""
 
     output_type = config["Variables_of_interest"]["type"]
-    if hasattr(data, "y_loc"):
+    if config["Training"]["compute_grad_energy"]:
```
pzhanggit, line 205: The changes look good to me. 
Could you merge these with lines 229-235 which also get `dims_list` from `config` for a different scenario, and turn off the Draft mode for a ci test run? Thanks.

RylieWeaver: Can you elaborate more on what you mean by margining with the lines 229-235?

pzhanggit: I was thinking if you could simplify the "if elif else" to "if else" since the new changes and the 229-235 are both about `dims_list = config["Variables_of_interest"]["output_dim"]`. Something like:

```
dims_list = config["Variables_of_interest"]["output_dim"]
if hasattr(data, "y_loc") and not config["Training"]["compute_grad_energy"]:
        for ihead in range(len(output_type)):
           < ...>
            assert dims_list[ihead]==dim_item
else:
      for ihead in range(len(output_type)):
            if output_type[ihead] != "graph":
                raise ValueError(
                    "y_loc is needed for outputs that are not at graph levels",
                    output_type[ihead],
                )


```

RylieWeaver: @pzhanggit

I think this will cause an error with the assert, because the output head will be of node type for `compute_grad_energy`.

pzhanggit: Yes, but also fixable? Anyway, what I proposed is a small suggestion that defines the `dims_list` from users input directly and keeps everything inside "if..else..." as a sanity checking for case setting up.
Feel free to keep the original changes. They look fine to me.

RylieWeaver: @allaffa

This PR is ready for your review now. In brief, it fixes the failure case of compute_grad_energy and removes the check.

In detail, the problem arises the following way:
```
1: Load dataset with ADIOS or Pickle:  calls update_config
2: update_config:  calls update_predicted_values() and update_config_NN_outputs()
   2.1: update_predicted_values():  defines y_loc according to node_feature_dims
   2.2: update_config_NN_outputs():  since y_loc exists, define output_dim by y_loc
3: Base() --> MLPNode():  defines output dimension of nodal head by output_dim
4: Conclusion:   When loading from ADIOS and Pickle, the nodal head output dimension is defined by node_feature_dim, **_not_** the output dimension defined in the config. This is a problem since we want the energy prediction to always be of size 1.
```

The proposed fix is:  If `compute_grad_energy`=true, then define output_dim according to the config in `update_config_NN_outputs()`, not according to y_loc.

allaffa:
@RylieWeaver 
Is this PR still current or is it now a duplicate due to these changes already being included in another PR?

allaffa:
@pzhanggit 
would you mind giving a last review to this PR?
Thanks

pzhanggit (approved):

PR #332, opened 2025-02-25
Smiles strings 321

kshitij-v-mehta:
Adds the SMILES string for the data object to ADIOS output


jychoi-hpc (approved):
Thank you for the update. It looks good to me.

allaffa (approved):

PR #337, opened 2025-03-31
remove pyg-lib requirement

allaffa:
Please confirm that indeed pyg-lib is useless


RylieWeaver (approved):
I approve these changes.

PyG-lib is made to speed up torch-geometric operations, so the only concern may be that removing PyG-lib will slow things down. However, given that:
(1) It's made specifically for NVIDIA/CUDA, while Frontier is AMD/ROCM
(2) The documentation is sparse
(3) It is a big hassle to include in conda

The drawbacks seem to outweigh the advantages and I approve.

PR #338, opened 2025-04-15
Dev for SC25 work on multi-level MTL in branch: Predictive gfm 2025 demo

pzhanggit:


jychoi-hpc (approved):
Thank you this PR. I checked a few scripts on Frontier (SC25-job-strong.sh, SC25-job-weak.sh, SC-baseline.sh, etc) and they ran ok. I think it is ok to merge. We can merge minor things after merging (e.g., `tmp_dict` in abstractbasedaset.py).

kshitij-v-mehta (changes requested):
We should not add scripts (SC25*.sh) to the top-level directory in the main branch. Can they be moved into a separate directory called `run-scripts/` (or something like that)?

pzhanggit:
> We should not add scripts (SC25*.sh) to the top-level directory in the main branch. Can they be moved into a separate directory called `run-scripts/` (or something like that)?

will do

kshitij-v-mehta:
Thanks, Pei. Could you also move the job-*.sh scripts into the new directory?

pzhanggit:
> Thanks, Pei. Could you also move the job-*.sh scripts into the new directory?

Oops, I missed those. Just moved them into new folder now. Thanks @kshitij-v-mehta

kshitij-v-mehta (approved):

PR #344, opened 2025-06-24
model wrapper to implicitly handle ipex optimizer wrapper for Intel GPUs backend

allaffa:


allaffa:
@jychoi-hpc 
I need to discuss with you how to handle the MultiTaskMLP model

jychoi-hpc:
@allaffa Sorry for my late reply. I am testing this PR now. We need this PR for Aurora. I will let you know soon how it goes.

jychoi-hpc:
@allaffa This PR mostly works. I just added my changes and created a PR for you to review: https://github.com/allaffa/HydraGNN/pull/44
I tested running on Aurora too. Once merged, it should be ok to merge to the main.

examples/multibranch/train.py
```diff
@@ -458,23 +458,27 @@ def info(*args, logtype="info", sep=" "):
         verbosity=verbosity,
     )
 
+    learning_rate = config["NeuralNetwork"]["Training"]["Optimizer"]["learning_rate"]
+
     ## task parallel
     if args.task_parallel:
         model = MultiTaskModelMP(model, branch_id, branch_group)
```
allaffa: @jychoi-hpc 

For aurora, we need to call the Intel optimize wrapper before calling the DDP wrapper. 
This PR already handles correctly this order when the model fits entire on one GPU. 
However, then the model is distributed across GPUs with multi-task parallelism, we discussed that we needed to be more careful about how we do this. 

I do not think that this PR handles correctly the use of `ipex.optimize` when Multi-Task parallelism is used.

jychoi-hpc: Missed in the update. I am going to add.

jychoi-hpc:
Just pushed the fix. Basically, in TaskParallel, it is ok to create the optimizer after the MultiTask model creation:
```
model = MultiTaskModelMP(model, branch_id, branch_group)
optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate)
model, optimizer = ipex.optimize(model, optimizer=optimizer)
```

`model.parameters()` returns only the parameters that is nothing to do with DDP.  Thus, we can call `ipex.optimize(...)` after model creation too.

jychoi-hpc:
Changed to make sure to follow the ALCF recommendation. I checked working on Aurora too.

jychoi-hpc (approved):

examples/multibranch/train.py
```diff
@@ -458,23 +464,29 @@ def info(*args, logtype="info", sep=" "):
         verbosity=verbosity,
     )
 
+    learning_rate = config["NeuralNetwork"]["Training"]["Optimizer"]["learning_rate"]
+
     ## task parallel
     if args.task_parallel:
+        optimizer = torch.optim.AdamW(model.parameters(), lr=learning_rate)
+        if hasattr(torch, "xpu") and torch.xpu.is_available():
```
allaffa, line 472: @jychoi-hpc 
For safety, I think it is best if this line and the next one come after the definition of the scheduler.

PR #345, opened 2025-07-21
Make SchNet RadGraph divert to CPU on Hardwares that don't have PyG GPU Support

RylieWeaver:
If `torch.cuda.is_available()`, use the radius graph as before. If not, then use a subclass of SchNet's radgraph that transfers tensors to CPU, executes the radgraph, and then puts the tensors back on the original device. 

The `torch.cuda.is_available()` check will be true for both CUDA and ROCm, so it still uses the GPU on Frontier as we'd want.


allaffa (approved):

allaffa:
@jychoi-hpc 
Let me know if this fix works for you

jychoi-hpc (approved):
I just tested on Aurora. It works well. @RylieWeaver thank you for the quick fix.

PR #346, opened 2025-07-21
Update ani1_x train.py

ArCho48:
Rectified spelling from self.max_neighbors to self.max_neighbours


ArCho48:
@allaffa I have added the --user-agent="Mozilla" code to ani1x, transition1x and mptrj download scripts. qm7x did not require this code and alexandria didnt work even with this code for reasons we discussed earlier. So, those I havent made any changes in those two dataset download scripts. This completes my commits

allaffa (approved):

PR #392, opened 2025-11-04
restructuring of handler for variables of interest

allaffa:
Restructure Variables of Interest Configuration System

This commit introduces a new self-documenting feature configuration format
for `Variables_of_interest` while maintaining full backward compatibility.

Key Changes:
- New self-documenting JSON format with explicit roles
- Automatic feature parsing and validation
- Removed ~200+ lines of redundant code
- Added comprehensive validation utilities

Core Components:
- `feature_config.py` (NEW)
- `config_utils.py` (MODIFIED)

Examples Migrated: 16
Test Configs Migrated: 5

Testing:
✅ All validation tests passed
✅ Legacy format backward compatibility verified
✅ Code formatted with black
✅ No breaking changes

Related: #392


USER_MANUAL.md
```diff
@@ -399,6 +476,72 @@ from hydragnn.utils.model import print_model
...
+```
+============================================================
+Feature Configuration Summary
+============================================================
+
+Node Features:
+------------------------------------------------------------
+  [0] atomic_number                dim= 1  role=input
+  [1] cartesian_coordinates        dim= 3  role=input
+  [2] forces                       dim= 3  role=internal
```
RylieWeaver: I think [2] forces is meant to say `role=output` here.

USER_MANUAL.md
```diff
@@ -332,19 +332,96 @@ HydraGNN provides extensive configuration options for building graph neural netw
...
 }
 ```
 
+**Feature Properties:**
+- `dim` (required): Integer dimension of the feature
+- `role` (required): One of:
+  - `"input"`: Feature used as model input
+  - `"output"`: Feature used as prediction target
+  - `"both"`: Feature used as both input and output
+- `output_type` (required for outputs): Either `"node"` or `"graph"`
```
RylieWeaver: I think that this is optional because there are lines like `result["type"].append(feat_config.get("output_type", "node"))` that set a default for the feature type.

hydragnn/utils/input_config_parsing/feature_config.py
```diff
@@ -0,0 +1,443 @@
...
+        "graph_feature_dims": [],
+        "input_node_features": [],
+        "output_names": [],
+        "output_index": [],
+        "output_dim": [],
+        "type": [],
+    }
+
+    # Parse node features
+    if "node_features" in var_config:
```
RylieWeaver, line 127: I believe that for this to work, the node/graph features within the data _**must**_ in the same order as in the config. I think this is a reasonable assumption/requirement, but it may be worth noting in the documentation.

This comment refers to the `_parse_new_format()` function iterates through the features; the code given above is not very informative

RylieWeaver (changes requested):
@allaffa

I agree that something like this could be very helpful to new users getting acclimated to HydraGNN. I noted some changes for documentation, but did not find any objectionable things that would cause errors.

hydragnn/utils/input_config_parsing/feature_config.py
allaffa: @RylieWeaver 
Thanks. I think I addressed this comment.

USER_MANUAL.md
allaffa: @RylieWeaver 
You are correct. I changed the documentation accordingly.

USER_MANUAL.md
allaffa: @RylieWeaver 
Thanks. I corrected it.

RylieWeaver (approved):

hydragnn/utils/input_config_parsing/feature_config.py
```diff
@@ -332,6 +332,29 @@ def _validate_against_data(parsed: dict, data_object) -> List[str]:
     return errors
 
 
+def validate_node_feature_columns(data_x, node_feature_dims, column_indices=None):
```
allaffa, line 332: @RylieWeaver 

I added this function to handle the edge case when (possibly) `data.x` contains more columns than the ones extracted. If you do not specify `column_index` for each feature, the number of columns in `data.x` must exactly match the sum of the dimensions of the node features specified in your configuration JSON, and the columns must be contiguous and in the order given. If you do specify `column_index` for each feature, `data.x` can have extra columns, and only the specified columns will be used for each feature.

RylieWeaver: @allaffa 

Would you envision calling this before or after the atom features are extracted with `extract_atom_features()` in `hydragnn/preprocess/graph_samples_checks_and_updates.py`? Currently I only see it in tests and an import.

hydragnn/utils/input_config_parsing/feature_config.py
```diff
@@ -0,0 +1,472 @@
...
+    # Parse node features
+    if "node_features" in var_config:
+        node_idx = 0
+        for feat_name, feat_config in var_config["node_features"].items():
+            result["node_feature_names"].append(feat_name)
+            result["node_feature_dims"].append(feat_config["dim"])
+
+            # Track input features
+            role = feat_config.get("role", "input")
+            if role in ["input", "both"]:
```
RylieWeaver: @allaffa 

Is there a scenario where we would want a feature to be both an input and an output?

hydragnn/utils/input_config_parsing/feature_config.py
```diff
@@ -0,0 +1,472 @@
...
+            node_idx += 1
+
+    # Parse graph features
+    if "graph_features" in var_config:
+        graph_idx = 0
+        for feat_name, feat_config in var_config["graph_features"].items():
+            result["graph_feature_names"].append(feat_name)
+            result["graph_feature_dims"].append(feat_config["dim"])
+
+            # Graph features are typically outputs (unless role explicitly set)
```
RylieWeaver: It may be beneficial to remove 'typically' here. I don't think there are any ways for a graph feature to serve as an input with how the architecture is currently set up.

hydragnn/utils/input_config_parsing/feature_config.py
```diff
@@ -0,0 +1,472 @@
...
+                result["input_node_features"].append(node_idx)
+
+            # Track output features
+            if role in ["output", "both"]:
+                result["output_names"].append(feat_name)
+                result["output_index"].append(node_idx)
+                result["output_dim"].append(feat_config["dim"])
+                result["type"].append(feat_config.get("output_type", "node"))
+
+            node_idx += 1
```
RylieWeaver, line 147: On a second look I think there may be something to fix here. I am not fully sure, so I will lay out my complete thinking so you can check it:

Assume we are using the automatically created config as you have added here
(1) Because of the for loop above, `node_idx` is appended to `result["input_node_features"]` for every node feature given in the config. So, for `n` provided features, then `result["input_node_features"] = [0, 1, ..., n-1]` no matter the dimensions of each of those node features.
(2) Later, in `hydragnn/preprocess/serialized_dataset_loader.py`, we have `self.input_node_features = config["NeuralNetwork"]["Variables_of_interest"]["input_node_features"]` --> `update_atom_features(self.input_node_features, data)`
(3) Then, in `hydragnn/preprocess/graph_samples_check_and_updates.py`, we have
```
def update_atom_features(atom_features: [AtomFeatures], data: Data):
```

In summary, I think this might be extracting only `n` features for `n` input node features, regardless of their respective dimensionalities (not what we want).

hydragnn/utils/input_config_parsing/feature_config.py
allaffa: @RylieWeaver 
Thank you for your thorough revision!
I added a new [commit ](https://github.com/ORNL/HydraGNN/pull/392/commits/8dce99f4f1b80adc2eb75b4f6065ee40de8f0dd3) where `update_predicted_values` now uses this check.

hydragnn/utils/input_config_parsing/feature_config.py
allaffa: @RylieWeaver 
Yes. There could be self-supervised learning (autoregressive) tasks where the same feature can be used as input and output. But I do also acknowledge that currently HydraGNN does not allow this use because `update_predicted_values` specifically extracts target nodal labels from `data.x` and moves them into `data.y`. For now, I removed this option in this [commit](https://github.com/ORNL/HydraGNN/pull/392/commits/66a2455548b7aa2f397d0fd93a9a05cbad863fb1)

hydragnn/utils/input_config_parsing/feature_config.py
allaffa: @RylieWeaver 
And you are correct here too. 
I tried to address this situation in this [commit](https://github.com/ORNL/HydraGNN/pull/392/commits/cd87a68b8125367333c2a3ff0a25017eb4c80cd3)

hydragnn/utils/input_config_parsing/feature_config.py
allaffa: @RylieWeaver 
Thanks. I adjusted the comment in this [commit](https://github.com/ORNL/HydraGNN/pull/392/commits/69001772aa33cd3c10026b1c752e1bfaf2995b3a)

hydragnn/utils/input_config_parsing/feature_config.py
RylieWeaver: @allaffa 

Nice! I think this will work, but consider putting it inside `update_atom_features`. 

The reason being this check _must_ be called before the node inputs are extracted and assigned to `data.x`, otherwise it will be checking for all the node features when only the node inputs are present. This has been done successfully so far because `update_predicted_values` is prior to `update_atom_features` everywhere I can find them both (serialized_dataloader, adiosdataset, distdataset, pickledataset), but it may be more robust to put it directly beforehand in the same function.

PR #501, opened 2026-06-08
Set CUDA device before NCCL initialization for distributed GFM runs

sparticlesteve:
## Title

Set CUDA device before NCCL initialization for distributed GFM runs

## Summary

- Add a `--datadir` option to `examples/multidataset_hpo_sc26/gfm_mlip_all_mpnn.py` so benchmark/data locations do not have to live under the example directory.
- Set the local CUDA device before initializing the NCCL process group. This ensures each rank has a deterministic current device during distributed initialization.

## Motivation

On Perlmutter, the GFM MLIP multidataset workflow can hang during FSDP v2/NCCL setup when ranks initialize distributed communication before selecting their local CUDA device. The benchmark wrapper can provide the data location, but the example script needs a `--datadir` argument to make that usable without editing source paths.

## Testing

Tested with the AmSC HydraGNN Perlmutter benchmark wrapper, using 1 GPU node, 4 ranks, Alexandria data under `/global/cfs/cdirs/amsc013/ccasert/hydragnn_data/dataset`, and a short `MAX_NUM_BATCH=1` run.

Relevant run configuration:

```bash
sbatch -A <gpu_account> \
  --export=ALL,HYDRAGNN_USE_FSDP=1,HYDRAGNN_FSDP_VERSION=2,DATA_DIR=/global/cfs/cdirs/amsc013/ccasert/hydragnn_data/dataset,MAX_NUM_BATCH=1 \
  run_perlmutter.sh
```

The benchmark wrapper sets `TRITON_CACHE_DIR=$SCRATCH/cache/triton` by default to avoid DeepSpeed/Triton cache shutdown hangs from home-directory cache locking.

Observed behavior:

- With the data-dir support but without the CUDA-device selection fix, the FSDP v2 run hangs during distributed/FSDP setup.
- With this change, the same FSDP v2 run trains for the short two-epoch smoke configuration and exits cleanly.

The Perlmutter-specific wrapper and environment details are maintained in the AmSC benchmarks repository; this PR keeps the HydraGNN changes limited to the portable example/data path and distributed initialization behavior.

### DIFF
--- a/examples/multidataset_hpo_sc26/gfm_mlip_all_mpnn.py
+++ b/examples/multidataset_hpo_sc26/gfm_mlip_all_mpnn.py
@@ -178,7 +178,12 @@ if __name__ == "__main__":
     parser.add_argument("--correlation", type=int, help="correlation", default=None)
     parser.add_argument("--nvme", action="store_true", help="use NVME")
     parser.add_argument("--startfrom", type=str, help="startfrom", default=None)
-    parser.add_argument("--datadir", type=str, help="path to directory containing .bp dataset files", default=None)
+    parser.add_argument(
+        "--datadir",
+        type=str,
+        help="path to directory containing .bp dataset files",
+        default=None,
+    )
 
     group = parser.add_mutually_exclusive_group()
     group.add_argument(
@@ -211,7 +216,9 @@ if __name__ == "__main__":
     node_feature_names = ["atomic_number", "cartesian_coordinates", "forces"]
     node_feature_dims = [1, 3, 3]
     dirpwd = os.path.dirname(os.path.abspath(__file__))
-    datadir = args.datadir if args.datadir is not None else os.path.join(dirpwd, "dataset")
+    datadir = (
+        args.datadir if args.datadir is not None else os.path.join(dirpwd, "dataset")
+    )
     ##################################################################################################################
     input_filename = os.path.join(dirpwd, args.inputfile)
     ##################################################################################################################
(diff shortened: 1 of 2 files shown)


sparticlesteve:
I'm preparing the draft MR on the amsc benchmark side which will include the actual job script for Max to review.

sparticlesteve:
Perlmutter testing details are captured in the AmSC benchmarks draft MR here: https://gitlab.com/amsc2/ai-services/at-scale-services/amsc-atscale-benchmarks/-/merge_requests/16

The relevant run script is:
https://gitlab.com/amsc2/ai-services/at-scale-services/amsc-atscale-benchmarks/-/blob/hydragnn/perlmutter-benchmark/benchmarks/training/hydragnn/run_perlmutter.sh

That wrapper sets the data directory via `--datadir`, enables FSDP through environment variables, and defaults `TRITON_CACHE_DIR` to `/pscratch/sd/s/sfarrell/cache/triton` to avoid DeepSpeed/Triton shutdown hangs from a home-directory cache.

sparticlesteve:
I added a follow-up commit (`43611e5`) to address the previous CI formatting failure from the `black . && git diff --exit-code` check. The change is formatting-only for the `--datadir` argument handling; the substantive changes remain the `--datadir` option and setting the local CUDA device before NCCL initialization.

allaffa (approved):

allaffa:
@jychoi-hpc 
I think this looks good. What do you think?

jychoi-hpc (approved):
It looks good to me.

PR #524, opened 2026-08-20
Add cost-aware graph batching

allaffa:
## Summary

- add a reusable model-independent `CostAwareBatchSampler`
- support deterministic node-budget packing with optional graph-count limits
- provide explicit error, singleton, and skip policies for oversized samples
- expose packing diagnostics and precomputed-cost support
- consume DDStore global shape metadata without reading graph payloads
- integrate `NeuralNetwork.Training.Batching` with HydraGNN data loaders
- distribute complete batches with equal step counts across ranks
- group similar-cost batches per synchronized distributed step
- rotate rank assignments and support deterministic padding or dropping
- update training batch counting and epoch propagation for batch samplers
- document configuration and add focused tests

## Configuration

```json
"Batching": {
  "mode": "node_budget",
  "max_nodes": 4096,
  "max_graphs": 64,
  "oversized_sample": "error",
  "shuffle": true,
  "seed": 0
}
```

## Scope

This PR implements both core batching and distributed cost-balanced rank
allocation. Dynamic online inference batching is intentionally reserved for a
separate follow-up PR because it addresses request serving rather than dataset
loading.

## Validation

- `pytest -q tests/test_cost_aware_batch_sampler.py tests/test_datasetclass_inheritance.py tests/test_config.py tests/test_precision_control.py`
- 15 passed, 1 skipped
- Python compile checks and `git diff --check`

### DIFF
--- a/hydragnn/train/train_validate_test.py
+++ b/hydragnn/train/train_validate_test.py
@@ -341,10 +341,12 @@ def train_validate_test(
         profiler.set_current_epoch(epoch)
         for dataloader in [train_loader, val_loader, test_loader]:
             sampler = getattr(dataloader, "batch_sampler", None)
-            if getattr(sampler, "set_epoch", None) is None:
-                sampler = dataloader.sampler
             if getattr(sampler, "set_epoch", None) is not None:
                 sampler.set_epoch(epoch)
+            else:
+                sampler = dataloader.sampler
+                if getattr(sampler, "set_epoch", None) is not None:
+                    sampler.set_epoch(epoch)
 
         with profiler as prof:
             tr.enable()
--- a/hydragnn/train/train_validate_test.py
+++ b/hydragnn/train/train_validate_test.py
@@ -171,11 +171,8 @@ def _set_reshard_after_backward(model, enabled):
 
 
 def get_nbatch(loader):
-    ## calculate numbrer of batches for a given loader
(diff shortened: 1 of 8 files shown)


hydragnn/train/train_validate_test.py
```diff
@@ -343,8 +340,11 @@ def train_validate_test(
         t0 = time.time()
         profiler.set_current_epoch(epoch)
         for dataloader in [train_loader, val_loader, test_loader]:
-            if getattr(dataloader.sampler, "set_epoch", None) is not None:
-                dataloader.sampler.set_epoch(epoch)
+            sampler = getattr(dataloader, "batch_sampler", None)
+            if getattr(sampler, "set_epoch", None) is None:
+                sampler = dataloader.sampler
+            if getattr(sampler, "set_epoch", None) is not None:
+                sampler.set_epoch(epoch)
```
jychoi-hpc, line 345: Maybe, is it better to use if-else?

jychoi-hpc (approved):
It looks great. I think you can ignore CoPilot comments on "pytest".

hydragnn/train/train_validate_test.py
allaffa: @jychoi-hpc 
Thanks. I changed it

